PR #3, opened 2019-05-22
Group By

ejoebstl:
Added
* group by support 
* aggregate_to_string aggregation function
* tests


src/ast/grouping.rs
```diff
@@ -0,0 +1,76 @@
...
+pub trait Groupable
+where
+    Self: Sized,
+{
+    /// Group by `self`
+    fn group(self) -> GroupByDefinition;
+}
+
+/// Convert the value into a group by definition.
+pub trait IntoGroupByDefinition {
```
pimeys: Why not using `From`?

```rust
impl<'a> From<&'a str> for GroupByDefinition {
    fn from(self) -> GroupByDefinition { ... }
}
```

src/ast/grouping.rs
```diff
@@ -0,0 +1,76 @@
+use crate::ast::{Column, DatabaseValue};
+
+pub type GroupByDefinition = (DatabaseValue);
```
pimeys: Why not just using `DatabaseValue` in `Grouping`? What does this type definition add to the code?

pimeys: Ah, now I get it. Limitation what can be used in the grouping!

src/ast/grouping.rs
ejoebstl: I was just replicating the pattern used for `IntoOrderDefinition`. Should I change to `From`?

pimeys: No, leave it how it is. I remember now there was some limitations in place here why we needed our own trait.

PR #4, opened 2019-05-23
WIP Connector Interfaces

ejoebstl:
PR to track database agnostic connector interfaces.

Databases to be ported from `prisma-rs`:

* [x] Sqlite
* [ ] Postgresql
* [ ] Mysql

WIP - Do not merge.


src/convenience.rs
```diff
@@ -0,0 +1,152 @@
+use std::collections::HashMap;
```
pimeys: I don't really see why this needs to be in prisma-query... And the naming of the file should be thought once more...

ejoebstl:
> The convenience methods are of no use in prisma, so feels like they should not stay in this crate.

Prisma is not only the query-engine, but also a load of other modules, in which it would be great to have this convenience.

Also, a by-name interface for query results should come with the DB connector, honestly. That's what I would expect form a DB connector. 

TBH, I'd rather remove the `convenience.rs` file and integrate the `ResultSet` struct tighter with the `Connection` and `Transaction` traits.

pimeys:
> > The convenience methods are of no use in prisma, so feels like they should not stay in this crate.
> 
> Prisma is not only the query-engine, but also a load of other modules, in which it would be great to have this convenience.
> 
> Also, a by-name interface for query results should come with the DB connector, honestly. That's what I would expect form a DB connector.
> 
> TBH, I'd rather remove the `convenience.rs` file and integrate the `ResultSet` struct tighter with the `Connection` and `Transaction` traits.

I'd be ok with integrating the `ResultSet`. Like we discussed, instead of using trait objects there, just allocate the rows once and then the columns. Use as little RAM as possible.

ejoebstl:
I've updated the PR according to our discussion.

The results/column names are now directly copied into the result set, which is then returned. 

I've also added a bunch of unit tests for the Sqlite connector interface.

I'm happy for feedback!

src/connector/sqlite.rs
```diff
@@ -0,0 +1,285 @@
...
+
+// Trait implementation for r2d2 sqlite.
+impl<'a> Connection for SqliteTransaction<'a> {
+    fn execute(&mut self, q: Query) -> QueryResult<Option<Id>> {
+        execute_impl(self, q)
+    }
+
+    fn query(&mut self, q: Query) -> QueryResult<ResultSet> {
+        query_impl(self, q)
+    }
```
pimeys: whitespace

src/connector/sqlite.rs
```diff
@@ -0,0 +1,285 @@
...
+
+// Trait implementation for r2d2 pooled connection, needed for RefCell.
+impl Connection for PooledConnection {
+    fn execute(&mut self, q: Query) -> QueryResult<Option<Id>> {
+        execute_impl(self, q)
+    }
+
+    fn query(&mut self, q: Query) -> QueryResult<ResultSet> {
+        query_impl(self, q)
+    }
```
pimeys: whitespace

src/connector/sqlite.rs
```diff
@@ -0,0 +1,285 @@
...
+}
+
+impl Connection for SqliteConnection {
+    fn execute(&mut self, q: Query) -> QueryResult<Option<Id>> {
+        execute_impl(self, q)
+    }
+
+    fn query(&mut self, q: Query) -> QueryResult<ResultSet> {
+        query_impl(self, q)
+    }
```
pimeys: whitespace

src/connector/sqlite.rs
```diff
@@ -0,0 +1,285 @@
...
+    params: &[ParameterizedValue],
+) -> QueryResult<ResultSet> {
+    let mut stmt = conn.prepare_cached(sql)?;
+    let mut rows = stmt.query(params)?;
+
+    let mut result = ResultSet::empty();
+
+    while let Some(row) = rows.next()? {
+        if result.name_to_index.len() == 0 {
+            result.name_to_index = ResultSet::build_name_map(&row.to_column_names())
```
pimeys: You can do this directly from `Rows` and don't need to call `len` for every row.

https://docs.rs/rusqlite/0.18.0/rusqlite/struct.Rows.html#method.column_names

pimeys: This could also be optional. If we have lots of columns and we don't need the name map, doing this wastes resources.

src/result_set.rs
```diff
@@ -0,0 +1,160 @@
...
+    /// Finds a column index for a name.
+    pub fn column_index(&self, name: &str) -> Result<usize, Error> {
+        match self.name_to_index.get(name) {
+            None => Err(Error::ColumnNotFound(String::from(name))),
+            Some(idx) => Ok(*idx),
+        }
+    }
+
+    /// Returns an interator over wrapped result rows.
+    pub fn iter(&self) -> impl std::iter::Iterator<Item = ResultRowWithName> {
```
pimeys: Not like this. Implement trait `IntoIterator<Item = ResultRowWithName>` and `Iterator<Item = ResultRowWithName>`.

https://doc.rust-lang.org/std/iter/trait.IntoIterator.html
https://doc.rust-lang.org/std/iter/trait.Iterator.html

src/result_set.rs
```diff
@@ -0,0 +1,160 @@
...
+    /// Creates a new instance, bound to the given column names and result rows.
+    pub fn new(names: &ColumnNames, values: Vec<ResultRow>) -> ResultSet {
+        ResultSet {
+            name_to_index: Self::build_name_map(names),
+            rows: values,
+        }
+    }
+
+    /// Creates a lookup map for column names.
+    pub(crate) fn build_name_map(names: &ColumnNames) -> HashMap<String, usize> {
```
pimeys: I'd say the `HashMap` should not be leaked out but kept as an internal implementation detail. Even though it's only public inside the crate.

src/connector/sqlite.rs
ejoebstl: Suggestion: We make loading the names lazy if I manage to pull the iterator/stream implementation of.

ejoebstl:
I've iterated on the feedback.

PR #8, opened 2019-06-12
Make Connectional trait object safe

mavilein:
This PR makes the trait `Connectional` object safe meaning it was not possible to use a `Box<Connectional>` or `Arc<Connectional>`. It was not object safe before as it used generic type parameters in its one method `with_connection`. This PR uses techniques described in `rustc --explain E0038` to fix this issue.

The resulting changes of this fix can be seen here: https://github.com/prisma/prisma/pull/4656

### DIFF
--- a/src/connector/mysql.rs
+++ b/src/connector/mysql.rs
@@ -58,6 +58,23 @@ impl Connectional for Mysql {
         let result = f(&mut conn);
         result
     }
+
+    fn execute_on_connection(&self, db: &str, query: Query) -> QueryResult<Option<Id>> {
+        self.with_connection(&db, |conn| conn.execute(query))
+    }
+
+    fn query_on_connection(&self, db: &str, query: Query) -> QueryResult<ResultSet> {
+        self.with_connection(&db, |conn| conn.query(query))
+    }
+
+    fn query_on_raw_connection(
+        &self,
+        db: &str,
+        sql: &str,
+        params: &[ParameterizedValue],
+    ) -> QueryResult<ResultSet> {
+        self.with_connection(&db, |conn| conn.query_raw(&sql, &params))
+    }
 }
 
 fn conv_params(params: &[ParameterizedValue]) -> my::params::Params {
--- a/src/connector/mysql.rs
+++ b/src/connector/mysql.rs
@@ -52,6 +52,7 @@ impl Connectional for Mysql {
     fn with_connection<F, T>(&self, _db: &str, f: F) -> QueryResult<T>
     where
         F: FnOnce(&mut Connection) -> QueryResult<T>,
+        Self: Sized,
     {
         dbg!(self.pool.state());
         let mut conn = self.pool.get()?;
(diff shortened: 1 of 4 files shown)


src/transaction.rs
```diff
@@ -45,7 +45,19 @@ pub trait Connectional {
     /// or schema mutations.
     fn with_connection<F, T>(&self, db: &str, f: F) -> QueryResult<T>
     where
-        F: FnOnce(&mut Connection) -> QueryResult<T>;
+        F: FnOnce(&mut Connection) -> QueryResult<T>,
+        Self: Sized;
+
+    fn execute_on_connection(&self, db: &str, query: Query) -> QueryResult<Option<Id>>;
```
pimeys: All these three methods from here to the end of the trait can be implemented in the trait, so you don't need to repeat yourself in the different connectors. They all call the `with_connection`, which should be the only method required in the connectors.

src/connector/sqlite.rs
```diff
@@ -50,9 +50,27 @@ impl Connectional for Sqlite {
     fn with_connection<F, T>(&self, db: &str, f: F) -> QueryResult<T>
     where
         F: FnOnce(&mut Connection) -> QueryResult<T>,
+        Self: Sized,
     {
         self.with_connection_internal(db, |c| f(c.get_mut()))
     }
+
```
pimeys: From here to the end, implement these three only in the trait.

src/connector/postgres.rs
```diff
@@ -75,13 +75,31 @@ impl Connectional for PostgreSql {
...
         F: FnOnce(&mut Connection) -> QueryResult<T>,
+        Self: Sized,
     {
         // TODO: Select DB.
         self.with_connection_internal(|mut client| {
             let result = f(&mut client);
             result
         })
     }
+
```
pimeys: From here to the end, implement these three only in the trait.

src/connector/mysql.rs
```diff
@@ -52,12 +52,30 @@ impl Connectional for Mysql {
...
     where
         F: FnOnce(&mut Connection) -> QueryResult<T>,
+        Self: Sized,
     {
         dbg!(self.pool.state());
         let mut conn = self.pool.get()?;
         let result = f(&mut conn);
         result
     }
+
```
pimeys: From here to the end, implement these three only in the trait.

ejoebstl: @mavilein, why did you add the Sized trait? 
This will, if I'm not mistaken, imply that all concrete implementations will need to be available at compile time.

ejoebstl:
@pimeys I suggest leaving the deduplication (e.g. moving all impls to the trait) to me, but after prisma day.

src/connector/mysql.rs
ejoebstl: Ah, nevermind, understood. It's better than using `impl Connectional` everywhere.

mavilein:
@ejoebstl : Yes agreed. This is ok for now but we should revisit after Prisma day. Dom has also seen some things that could be improved. Let's do this as a team after Prisma day.

PR #12, opened 2019-07-27
ParameterizedValue: Add method into_vec

aknuds1:
Add method `ParameterizedValue::into_vec`.

TODO: 

- [x] Use conversion methods in `TryFrom` implementations.

### DIFF
--- a/src/ast/values.rs
+++ b/src/ast/values.rs
@@ -1,5 +1,7 @@
 use crate::ast::*;
+use crate::error::Error;
 use std::borrow::{Borrow, Cow};
+use std::convert::TryFrom;
 
 #[cfg(feature = "json-1")]
 use serde_json::{Number, Value};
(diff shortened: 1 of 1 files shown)


src/ast/values.rs
```diff
@@ -267,6 +287,62 @@ impl<'a> From<i32> for ParameterizedValue<'a> {
...
+        }
+     }
+}
+
+impl<'a> TryFrom<ParameterizedValue<'a>> for f64 {
+    type Error = &'static str;
+
+    fn try_from(value: ParameterizedValue<'a>) -> Result<f64, Self::Error> {
+        match value {
+            ParameterizedValue::Real(v) => Ok(v),
```
pimeys: Maybe use the dedicated conversion methods from `ParameterizedValue` here and return an exact `Error` from this crate's error type.

PR #15, opened 2019-09-18
Support CAST expressions

tomhoule:
It's the portable way to cast SQL values between types.

Postgres: https://www.postgresql.org/docs/11/sql-expressions.html#SQL-SYNTAX-TYPE-CASTS
SQLite: https://www.sqlite.org/lang_expr.html#castexpr
MySQL: https://dev.mysql.com/doc/refman/5.7/en/cast-functions.html

The CI failure doesn't seem related.

### DIFF
--- a/src/visitor.rs
+++ b/src/visitor.rs
@@ -524,6 +524,11 @@ pub trait Visitor<'a> {
                     )
                 }
             }
+            FunctionType::Cast(cast) => format!(
+                "CAST({expr} AS {tpe})",
+                expr = self.visit_database_value(*cast.expr),
+                tpe = cast.tpe
+            ),
             FunctionType::Count(fun_count) => {
                 if fun_count.exprs.is_empty() {
                     String::from("COUNT(*)")
--- a/.buildkite/docker.sh
+++ b/.buildkite/docker.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+MYSQL_ROOT_PASSWORD=prisma
+
 docker network create test-net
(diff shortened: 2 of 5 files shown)


src/visitor.rs
```diff
@@ -524,6 +524,11 @@ pub trait Visitor<'a> {
                     )
                 }
             }
+            FunctionType::Cast(cast) => format!(
+                "CAST({expr} AS {tpe})",
+                expr = self.visit_database_value(*cast.expr),
+                tpe = cast.tpe
```
tomhoule: Not sure we manipulate the type as a raw string. I haven't looked extremely in depth but I haven't found a wrapper for types in the AST.

pimeys: Should you parameterize the `tpe` and did you try it out in a database query? Maybe using an enum for all possible cast types...

tomhoule:
The CI failures may be due to the MySQL version: the tests passed with the mysql 5.7 I tested with locally but fail on CI with `mysql:latest`.

I'll think a bit more about how to insert the type in a safer way. It seems that at least in postgres, passing a type as a parameter doesn't work:

```
postgres=# PREPARE prep1 AS SELECT CAST($1 AS $2);
ERROR:  syntax error at or near "$2"
LINE 1: PREPARE prep1 AS SELECT CAST($1 AS $2);
```

We could still have an enum with the supported types to introduce some safety there.

tomhoule:
Closing this because problems surfaced when writing tests, and this is not needed immediately.

PR #17, opened 2019-10-15
Expose TLS-related errors separately from other connection errors

tomhoule:
This is meant to enable better error diagnostics in Lift. This also means we now have to handle both cases in the rest of prisma-engine, so I'd like some feedback on that.


tomhoule:
CI failure seems unrelated.

PR #23, opened 2019-11-01
Limit allocations when visiting queries

pimeys:


src/visitor.rs
```diff
@@ -358,209 +396,243 @@ pub trait Visitor<'a> {
...
     }
 
-    fn visit_partitioning(&mut self, over: Over<'a>) -> String {
-        let mut result = Vec::new();
-
+    fn visit_partitioning(&mut self, over: Over<'a>) -> fmt::Result {
         if !over.partitioning.is_empty() {
-            let mut parts = Vec::new();
+            let len = over.partitioning.len();
+            self.write("PARTITION BY ")?;
```
tomhoule: not positive, but wouldn't a space at the beginning of the string be safer here?

src/visitor/mysql.rs
```diff
@@ -20,52 +22,62 @@ impl<'a> Visitor<'a> for Mysql<'a> {
         Q: Into<Query<'a>>,
     {
         let mut mysql = Mysql {
-            parameters: Vec::new(),
+            query: String::with_capacity(4096),
+            parameters: Vec::with_capacity(1024),
```
tomhoule: We just discussed: this is probably too big for the common case.

src/visitor.rs
pimeys: no :(

PR #26, opened 2019-11-05
Expose postgres/mysql URL parsing logic

tomhoule:


pimeys:
Why not add methods to the rest of the connection parameters? The supported/unsupported is not necessary due to us not really sending the connection string to the crate. I'd like to see nice accessors for SSL, connection_limit etc.

tomhoule:
Good point - I added only those I needed, but better be exhaustive. I'll add these.

tomhoule:
@pimeys  done

tomhoule:
The new methods return structs to preserve the exhaustiveness check we had before (i.e. the logging of unknown parameters).

PR #28, opened 2019-11-08
Store SQLite file paths as String

tomhoule:
This is changed from PathBuf because we need the file paths to be valid
UTF-8 to pass them in as parameters to `ATTACH ? AS ?` queries. This
would result in repeated fallible conversions that are removed by this
commit.


tomhoule:
@pimeys I'm merging this without review because we went through it together.

PR #29, opened 2019-11-08
Handle more errors (MySQL, Postgres)

tomhoule:


tomhoule:
@pimeys I assume it's uncontroversial so I'm just going to merge this. It's kind of tested within the errors spec branch on prisma-engine, but I could spend time to write proper tests for this in quaint.

Issue #36, opened 2019-11-19
Database strings starting with `sqlite:` are not supported


tomhoule:
The parsing  logic for `SqliteParams` assumes `file:`.

PR #37, opened 2019-11-19
Trim sqlite: from the path if there

pimeys:


pimeys:
Fixes https://github.com/prisma/quaint/issues/36

PR #38, opened 2019-11-20
Remove params structs

tomhoule:


tomhoule:
I checked it doesn't impact prisma-engine. Merging this since we already discussed it.

PR #42, opened 2019-11-25
Switch to mobc pool

pimeys:
The author of mobc is super responsive and the library fits like a glove to Quaint. Now we also get rid of direct tokio dependency and can support async-std.


src/error.rs
```diff
@@ -67,14 +64,12 @@ pub enum Error {
...
-#[cfg(any(
-    feature = "mysql-16",
-    feature = "postgresql-0_16",
-    feature = "rusqlite-0_19"
-))]
-impl From<r2d2::Error> for Error {
-    fn from(e: r2d2::Error) -> Error {
-        Error::ConnectionError(e.into())
+impl From<mobc::Error<failure::Compat<Error>>> for Error {
+    fn from(e: mobc::Error<failure::Compat<Error>>) -> Self {
```
tomhoule: It's going to get nicer when we switch away from failure. Maybe it's something we should consider doing before 0.1?

src/lib.rs
```diff
@@ -215,29 +215,29 @@ impl Quaint {
...
+            .await?;
+
+        Ok(Self { inner, connection_info })
     }
 
     /// The number of connections in the pool.
-    pub fn capacity(&self) -> usize {
-        self.inner.capacity()
+    pub async fn capacity(&self) -> u32 {
+        self.inner.state().await.connections
```
tomhoule: it feels a bit weird that you have to await to get the pool's internal state, but it's fine

src/error.rs
pimeys: Naturally.

src/lib.rs
pimeys: The state is blocked by an async mutex.

tomhoule: makes sense

PR #46, opened 2019-12-02
Update to tokio 0.2 and futures 0.3

pimeys:


pimeys:
Dum dum. Need to add the new feature flags to buildkite.

src/connector/postgres.rs
```diff
@@ -85,12 +85,12 @@ impl SslParams {
         auth.accept_mode(self.ssl_accept_mode);
 
         if let Some(ref cert_file) = self.certificate_file {
-            let cert = fs::read(cert_file).await?;
+            let cert = fs::read(cert_file)?;
```
tomhoule: hmmm I don't know what the right thing to do would be here if we don't want to always depend on async-std, but reverting to blocking file IO feels bad

src/connector/connection_info.rs
```diff
@@ -260,6 +226,7 @@ impl fmt::Display for SqlFamily {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(feature = "sqlite")]
```
tomhoule: why only on sqlite?

Cargo.toml
```diff
@@ -17,11 +17,22 @@ keywords    = ["mysql", "postgresql", "sqlite", "sql"]
...
+
+full = ["pooled", "sqlite", "json-1", "postgresql", "uuid-0_7", "chrono-0_4", "mysql"]
+full-postgresql = ["pooled", "postgresql", "json-1", "uuid-0_7", "chrono-0_4", "array"]
+full-mysql = ["pooled", "mysql", "json-1", "uuid-0_7", "chrono-0_4"]
+full-sqlite = ["pooled", "sqlite", "json-1", "uuid-0_7", "chrono-0_4"]
+
+single = ["sqlite", "json-1", "postgresql", "uuid-0_7", "chrono-0_4", "mysql"]
+single-postgresql = ["postgresql", "json-1", "uuid-0_7", "chrono-0_4", "array"]
+single-mysql = ["mysql", "json-1", "uuid-0_7", "chrono-0_4"]
+single-sqlite = ["sqlite", "json-1", "uuid-0_7", "chrono-0_4"]
```
tomhoule: :100: I prefer named features, looks nice!

src/connector/postgres.rs
pimeys: I don't want to rely on a runtime that spawns random threads on connect. The facilities should come from the current runtime or we just block. More we rely on certain runtime functions, harder it is to support other runtimes.

src/connector/connection_info.rs
pimeys: When you compile without sqlite feature, this gives a warning.

Cargo.toml
pimeys: Yes. Some logic here finally.

PR #48, opened 2019-12-03
Flip postgres SSL certificate verification default

tomhoule:


tomhoule:
Merging this since we talked about it over coffee.

tomhoule:
I thought about it but decided to keep the current behaviour. The default is always `accept_invalid_certificates`, but if you provide an explicit mode and it doesn't work, it's safer to assume you wanted the strict behaviour. But maybe it hurts consistency too much, idk.

tomhoule:
Maybe it should be a hard error.

PR #61, opened 2020-01-14
Removing execute/execute_raw from queryable

pimeys:
We made an assumption earlier that an execute function should return an
`Id` in a case of an insert. We can fetch the last id from certain
databases, but it will only work with an auto-incremented integer field
using MySQL or in case of SQLite we get a row_id which might not be what
we think it is.

This change removes `execute` and `execute_raw` methods from `Queryable`.
Everything they did can be done using `query` and `query_raw`. If the
database gives us the last id, we store it to the `ResultSet` and in
case of PostgreSQL, the query should contain a `RETURNING` part to fetch
the inserted data.

This is a breaking change and should change Quaint version to 0.2.


pimeys:
Maybe I should add some...

Sent from [ProtonMail](https://protonmail.ch), encrypted email based in Switzerland.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #65, opened 2020-01-19
Replaces `name_to_index: BTreeMap<String, usize>` with `columns: Vec<String>`

KyleU:
This preserves column ordering. No public interface has changed. Construction should be faster, column name lookups are now linear


pimeys:
Wow nice thanks. I think I'll merge this to master and to the 0.1 branch, so it gets to the next 0.1 version.

PR #66, opened 2020-01-20
Timeouts

pimeys:


src/connector/sqlite.rs
```diff
@@ -99,10 +106,14 @@ impl TryFrom<&str> for Sqlite {
 
     fn try_from(path: &str) -> crate::Result<Self> {
         let params = SqliteParams::try_from(path)?;
-        let client = Mutex::new(rusqlite::Connection::open_in_memory()?);
+
+        let conn = rusqlite::Connection::open_in_memory()?;
+        conn.busy_timeout(params.socket_timeout)?;
```
tomhoule: on sqlite this will probably lead to blocking code in async context, so I'm a bit worried about it but I don't have an alternative, except doing it from outside of quaint or using async-std with the new executor for sqlite.

pimeys: I mean we already block the thread with SQLite anyhow. This will timeout all other queries that are waiting a lock to be released for too long. The one heavy blocking query is not affected though.

PR #67, opened 2020-01-21
Try to avoid prepared statements when starting Quaint

pimeys:


pimeys:
Merged manually to 0.1 and 0.2 branches.

PR #72, opened 2020-01-30
Errors should hold the original code and message

pimeys:
This is a breaking change. For whatever code that relied on matching the `Error` will now have to match the `ErrorKind` that's returned from `kind()`.

### DIFF
--- a/src/connector/postgres/error.rs
+++ b/src/connector/postgres/error.rs
@@ -1,4 +1,4 @@
-use crate::error::{DatabaseConstraint, Error};
+use crate::error::{DatabaseConstraint, Error, ErrorKind};
 
 impl From<tokio_postgres::error::Error> for Error {
     fn from(e: tokio_postgres::error::Error) -> Error {
@@ -6,7 +6,8 @@ impl From<tokio_postgres::error::Error> for Error {
 
         match e.code().map(|c| c.code()) {
             // Don't look at me, I'm hideous ;((
-            Some("23505") => {
+            Some(code) if code == "23505" => {
+                let code = code.to_string();
                 let error = e.into_source().unwrap(); // boom
                 let db_error = error.downcast_ref::<DbError>().unwrap(); // BOOM
                 let detail = db_error.detail().unwrap(); // KA-BOOM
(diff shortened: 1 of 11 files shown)


src/connector/postgres/error.rs
```diff
@@ -44,7 +54,12 @@ impl From<tokio_postgres::error::Error> for Error {
                 let splitted: Vec<&str> = splitted[1].split('"').collect();
                 let db_name = splitted[1].into();
 
-                Error::DatabaseDoesNotExist { db_name }
+                let mut builder = Error::builder(ErrorKind::DatabaseDoesNotExist { db_name });
+
+                builder.set_original_code("3D000");
```
tomhoule: it worries me a tiny bit that the error code is repeated here from the pattern, and they could get out of sync. It's less pretty, but the pattern could be `Some(code) if code = "3D000"`, then here it can be `builder.set_original_code(code)`. Just a suggestion though, I don't think it's too bad how it is.

src/error.rs
```diff
@@ -18,7 +18,73 @@ impl fmt::Display for DatabaseConstraint {
...
+    pub fn kind(&self) -> &ErrorKind {
+        &self.kind
+    }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.kind.fmt(f)
+    }
+}
```
tomhoule: no need to do it now, but we could include the original message and code (if present) in the error message too. But maybe that's confusing for people if they see that and expect prisma error codes.

src/error.rs
```diff
@@ -18,7 +18,73 @@ impl fmt::Display for DatabaseConstraint {
...
+impl ErrorBuilder {
+    pub(crate) fn set_original_code(&mut self, code: impl Into<String>) -> &mut Self {
+        self.original_code = Some(code.into());
+        self
+    }
+
+    pub(crate) fn set_original_message(&mut self, message: impl Into<String>) -> &mut Self {
+        self.original_message = Some(message.into());
+        self
+    }
```
tomhoule: hehe, the eternal dilemma of `Self` vs `&mut Self` builders

src/connector/postgres/error.rs
pimeys: hwut? done.

src/error.rs
pimeys: I don't really want that...

src/error.rs
pimeys: yeah... or no chaining at all?

PR #73, opened 2020-01-31
Enum support

do4gr:


src/connector/postgres/conversion.rs
```diff
@@ -283,6 +300,10 @@ impl<'a> ToSql for ParameterizedValue<'a> {
                 }
                 _ => float.to_sql(ty, out),
             },
+            ParameterizedValue::Enum(string) => {
+                out.extend_from_slice(format!("{}", string).as_bytes());
```
pimeys: Why you double allocate with `format`, don't you have the ownership to the string, so you can just call `as_bytes` directly?

src/connector/postgres/conversion.rs
```diff
@@ -322,6 +343,10 @@ impl<'a> ToSql for ParameterizedValue<'a> {
                 _ => float.to_sql(ty, out),
             },
             ParameterizedValue::Text(string) => string.to_sql_checked(ty, out),
+            ParameterizedValue::Enum(string) => {
+                out.extend_from_slice(format!("{}", string).as_bytes());
```
pimeys: Same as above.

src/connector/postgres/conversion.rs
do4gr: fixed

src/connector/postgres/conversion.rs
do4gr: fixed

PR #78, opened 2020-02-04
More Enums

do4gr:


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+
```
pimeys: We don't need this much of whitespace here...

src/connector/postgres/conversion.rs
```diff
@@ -221,10 +221,35 @@ impl GetRow for PostgresRow {
...
+                                        .collect(),
+                                )
+                            }
+                            None => ParameterizedValue::Null,
+                        },
+                        _ => match row.try_get(i)? {
+                            Some(val) => {
+                                let val: Vec<String> = val;
+                                ParameterizedValue::Array(
+                                    val.into_iter().map(|x| ParameterizedValue::Text(x.into())).collect(),
```
pimeys: Hmm, looking at this aren't we just converting everything that is not an enum to a string? What about integers? This smells like we could just call the function recursively with the inner kind to get a proper conversion...

do4gr: Yes, we could probably do that. But I do not want to do this now since I only need enum arrays. This just handles enums and the rest is handled like before.

PR #85, opened 2020-02-26
Add ParameterizedValue::IpAddress for postgres

tomhoule:


src/ast/values.rs
```diff
@@ -265,6 +271,19 @@ impl<'a> ParameterizedValue<'a> {
         }
     }
 
+    #[cfg(feature = "postgresql")]
+    pub fn is_ip_address(&self) -> bool {
+        self.as_ip_address().is_some()
+    }
+
+    #[cfg(feature = "postgresql")]
+    pub fn as_ip_address(&self) -> Option<&std::net::IpAddr> {
```
pimeys: `IpAddr` is `Copy` so return by value.

src/connector/mysql/conversion.rs
```diff
@@ -105,6 +105,8 @@ impl<'a> From<ParameterizedValue<'a>> for MyValue {
                 dt.second() as u8,
                 dt.timestamp_subsec_micros(),
             ),
+            #[cfg(feature = "postgresql")]
+            ParameterizedValue::IpAddress(ip_addr) => MyValue::Bytes(ip_addr.to_string().into_bytes()),
```
pimeys: hmm...

?

pimeys: I know this is kind of needed if you have both features enabled. But maybe the enum variant should not be under a feature flag then...

src/connector/sqlite/conversion.rs
```diff
@@ -68,6 +68,9 @@ impl<'a> ToSql for ParameterizedValue<'a> {
             ParameterizedValue::Uuid(value) => ToSqlOutput::from(value.to_hyphenated().to_string()),
             #[cfg(feature = "chrono-0_4")]
             ParameterizedValue::DateTime(value) => ToSqlOutput::from(value.timestamp_millis()),
+
+            #[cfg(feature = "postgresql")]
+            ParameterizedValue::IpAddress(ip_addr) => ToSqlOutput::from(ip_addr.to_string().into_bytes()),
```
pimeys: Same here. maybe do not hide the variant under a feature flag if all database connectors still need to react to the values if postgres is enabled.

pimeys:
Also what bothers me a bit is that with PostgreSQL when we read an IP address value from the database, it'll be an IP address. If then we read an IP address from MySQL or SQLite, it will not be an IP address variant from ParameterizedValue.

This smells like a problem with our architecture with parameterized values...

tomhoule:
Yes, maybe a trait-based type mapping approach like `tokio-postgres` would be more flexible than the enum.

pimeys:
Do you have an idea how that would look like in a multi-database context such as Quaint?

tomhoule:
We need to think really hard about this, the traits would have to be very generic.

Also we would probably have `Quaint<Postgres>`, `Quaint<Sqlite>` and so on to allow some types of parameters, etc. Then that type parameter would probably end up everywhere in the AST, it would be messy. But possible more powerful than an enum.

pimeys:
Yeah... I don't really like how we have database-specific enum variants now and especially how you cannot map them from and back from the database so that they'll always turn into the same variant.

pimeys:
Remember also how `ParameterizedValue` is the singular enum that will be parameterized out from the SQL queries. That role is extremely important, but maybe we should not return `ParameterizedValue`s in the `ResultSet`s.

tomhoule:
Closing this, since we went with not extending `ParameterizedValue` for now.

tomhoule:
Closing this for now, since we need to decide on an approach to have a rich type system for result sets in the future.

PR #90, opened 2020-03-05
Handle many more postgres types

tomhoule:


src/connector/postgres.rs
```diff
@@ -548,6 +548,78 @@ mod tests {
...
+                bytes_uuid uuid,
+                network_inet inet,
+                numeric_float4 float4,
+                numeric_float8 float8,
+                time_timetz timetz,
+                time_time time,
+                time_date date,
+                text_jsonb jsonb,
+                time_timestamptz timestamptz
+            );
```
tomhoule: we may be missing some here - I have more exhaustive tests in the query engine.

Also we're not testing arrays of these types, I'll need to confirm that they work.

PR #91, opened 2020-03-06
Add support for money columns on postgres

tomhoule:
We assume the default precision of 2 decimals here. If the server is
configured differently, the values will be wrong.

### DIFF
--- a/src/connector/postgres/conversion.rs
+++ b/src/connector/postgres/conversion.rs
@@ -417,10 +417,12 @@ impl<'a> ToSql for ParameterizedValue<'a> {
             }
             (ParameterizedValue::Real(decimal), &PostgresType::MONEY) => {
                 let mut i64_bytes: [u8; 8] = [0; 8];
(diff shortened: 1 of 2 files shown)


src/connector/postgres/conversion.rs
```diff
@@ -391,8 +416,13 @@ impl<'a> ToSql for ParameterizedValue<'a> {
                 f.to_sql(ty, out)
             }
             (ParameterizedValue::Real(decimal), &PostgresType::MONEY) => {
-                let f = decimal.to_f64().expect("decimal to f64 conversion");
-                f.to_sql(ty, out)
+                let mut decimal = decimal.clone();
+                // The default for money values on Postgres.
+                decimal.set_scale(2)?;
```
tomhoule: the `set_scale` is probably not necessary, in the end.

tomhoule:
Wait I think I got it wrong.

tomhoule:
This is ready for review now @pimeys

PR #94, opened 2020-03-09
Optimize certain IN selections

pimeys:
If we select with a row.in_selection(values), we want to optimize cases
where row and values both contain only one item.

So that

```sql
(foo) IN ((1), (2))
```

becomes

```sql
foo IN (1, 2)
```

### DIFF
--- a/src/ast/values.rs
+++ b/src/ast/values.rs
@@ -817,13 +817,19 @@ impl<'a> Values<'a> {
         Self::default()
     }
 
+    /// Create a new in-memory set of values with an allocated capacity.
+    pub fn with_capacity(capacity: usize) -> Self {
+        Self {
+            rows: Vec::with_capacity(capacity)
+        }
+    }
+
     /// Add value to the temporary table.
-    pub fn push<T>(mut self, row: T) -> Self
+    pub fn push<T>(&mut self, row: T)
     where
         T: Into<Row<'a>>,
     {
         self.rows.push(row.into());
-        self
     }
 
     /// The number of rows in the in-memory table.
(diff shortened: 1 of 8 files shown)


src/ast/values.rs
```diff
@@ -835,6 +841,26 @@ impl<'a> Values<'a> {
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
+    pub fn row_len(&self) -> usize {
+        match self.rows.split_first() {
+            Some((row, _)) => row.len(),
+            None => 0,
```
tomhoule: nice!

src/visitor.rs
```diff
@@ -487,35 +487,47 @@ pub trait Visitor<'a> {
...
-                    }
-                    dbv => {
-                        self.visit_database_value(*left)?;
-                        self.write(" IN ")?;
-                        self.visit_database_value(dbv)
-                    }
-                },
+            Compare::In(left, right) => match (*left, *right) {
+                (_, DatabaseValue::Row(ref row)) if row.is_empty() => self.write("1=0"),
+                (DatabaseValue::Row(mut cols), DatabaseValue::Values(vals)) if cols.len() == 1 && vals.row_len() == 1 => {
```
tomhoule: Isn't it an error if the lengths don't match here?

src/ast/values.rs
```diff
@@ -835,6 +841,26 @@ impl<'a> Values<'a> {
...
+
+        for mut row in self.rows.into_iter() {
+            match row.pop() {
+                Some(value) => result.push(value),
+                None => return None
+            }
+        }
+
+        Some(result)
+    }
```
tomhoule: nice!

PR #98, opened 2020-03-17
Fix and test array of bit vectors support on postgres

tomhoule:


tomhoule:
@pimeys  merging this right away because it's kind of trivial

PR #102, opened 2020-03-20
Set maximum connection idle time

pimeys:
Tries to address https://github.com/prisma/prisma2/issues/1559


pimeys:
FORGETABOUTIT. This is not a good idea, the connection will get kicked out even if in active use after the timeout.

PR #107, opened 2020-03-26
Pool builder

pimeys:
Allows starting a pool builder from an uri (that parses some of the
settings), then changing them easily afterwards. Also some new settings
allowed for the user:

- `max_idle_lifetime`
- `test_on_check_out`
- `health_check_interval`

### DIFF
--- a/src/pooled.rs
+++ b/src/pooled.rs
@@ -1,10 +1,108 @@
-//! A connection pool to a SQL database.
+//! # A connection pool to a SQL database.
+//!
+//! A pool is created through the [`builder`] method, starting from a connection
+//! string that allows some of the parameters be delivered by the user.
+//!
+//! A connection string has the following structure:
+//!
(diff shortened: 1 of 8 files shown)


src/pooled.rs
```diff
@@ -1,10 +1,108 @@
...
+//!
+//! ## Common parameters
+//!
+//! - `connection_limit` defines the number of connections opened to the
+//!   database.
+//!
+//! ## SQLite
+//!
+//! - `user`/`password` do not do anything and can be emitted.
+//! - `host` should point to the database file.
```
tomhoule: do we really parse it as a URL on sqlite? As far as I remember it was something we stopped doing because it didn't work nicely with some characters, including spaces, in file paths. Should we mention that in the docs?

pimeys: We don't use `Url` but we treat the string as one.

PR #114, opened 2020-04-02
Upgrade mysql_async to 0.23

tomhoule:
For the float fixes.

### DIFF
--- a/src/connector/mysql/conversion.rs
+++ b/src/connector/mysql/conversion.rs
@@ -32,6 +32,7 @@ impl TakeRow for my::Row {
                 // TOOD: This is unsafe
                 my::Value::UInt(i) => ParameterizedValue::Integer(i as i64),
                 my::Value::Float(f) => ParameterizedValue::from(f),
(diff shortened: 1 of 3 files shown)


src/connector/mysql/conversion.rs
```diff
@@ -89,7 +90,7 @@ impl<'a> From<ParameterizedValue<'a>> for MyValue {
         match pv {
             ParameterizedValue::Null => MyValue::NULL,
             ParameterizedValue::Integer(i) => MyValue::Int(i),
-            ParameterizedValue::Real(f) => MyValue::Float(f.to_f64().expect("Decimal is not a f64.")),
+            ParameterizedValue::Real(f) => MyValue::Double(f.to_f64().expect("Decimal is not a f64.")),
```
pimeys, line 93: What happens here if the underlying column is a float, we write a double and then read it back?

tomhoule: The database will cast it, most likely. I can write a test for it.

tomhoule: Test written.

PR #116, opened 2020-04-09
Convert MySQL NEWDECIMAL to a numeric value

tomhoule:
It is returned from aggregate functions like SUM.

Previously returned as bytes.


tomhoule:
Yes, I'm a bit afraid of that too. I'm going to have a look at all the constants for `ColumnType` and see if we may be missing something.

PR #121, opened 2020-04-27
Ast types revamp

pimeys:
`Expression` was almost nothing merged with `DatabaseValue`. `DatabaseValue` should be `Expression` and `ParameterizedValue` should be `Value`.

### DIFF
--- a/src/ast/expression.rs
+++ b/src/ast/expression.rs
@@ -1,18 +1,236 @@
 use crate::ast::*;
+use std::borrow::Cow;
 
-/// A database expression.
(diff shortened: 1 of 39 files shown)


src/ast/expression.rs
```diff
@@ -1,18 +1,236 @@
 use crate::ast::*;
+use std::borrow::Cow;
 
-/// A database expression.
-#[derive(Debug, PartialEq, Clone)]
+/// A value we can compare and use in database queries.
```
tomhoule, line 4: Should this say `expression`?

Issue #89, opened 2020-03-04
Assign an alias to a value column


scull7:
First,  thank you, the library is awesome.  

Second, I'm having difficulty understanding how to add an alias to an aggregate column. Specifically when using the `aggregate_to_string` function (as seen below).
    
```rust
    Select::from_table("employee")
        .column(("employee", "id"))
        .column(("employee", "created"))
        .column(("employee", "updated"))
        .column(("employee", "deleted"))
        .column(("employee", "casino_id"))
        .column(("employee", "uuid"))
        .column(("employee", "name_first"))
        .column(("employee", "name_last"))
        .column(("employee", "title"))
        .value(aggregate_to_string(Column::from(("role", "display"))))
        .inner_join(join_credential)
        .inner_join(join_credential_role)
        .inner_join(join_role)
        .group_by(Column::from(("employee", "id")))
```

Is it possible to assign an alias to the `.value(aggregate_to_string(Column::from(("role", "display"))))`?

pimeys:
Yeah. I was hit by the same problem once, and what I found out it would be nice if that function returns a `Function` type directly. How you can do it now is `Function::from(aggregate_to_string(Column::from(("role", "display")))).alias("foo")`. What kind of sucks in the current api is how the tuples are by-default table definitions, leading to the `Column::from` syntax, and how the function functions (sic) return something that is not a `Function` enum, and don't allow you to alias them directly.

This is something we should think about for the upcoming 0.2 release.

pimeys:
https://docs.rs/quaint/0.2.0-alpha.9/quaint/ast/struct.Function.html

pimeys:
Also some examples would be super cool before we kick the 0.2 out from the door. Ping @tomhoule let's do that soon :)

scull7:
Thank you for pointing me in the right direction.  I just was able to go back and change the code to use quaint instead of my SQL file. Worked beautifully.

```rust
Select::from_table("employee")
        .column(Column::from(("employee", "id")))
        .column(Column::from(("employee", "created")))
        .column(Column::from(("employee", "updated")))
        .column(Column::from(("employee", "deleted")))
        .column(Column::from(("employee", "casino_id")))
        .column(Column::from(("employee", "uuid")))
        .column(Column::from(("employee", "name_first")))
        .column(Column::from(("employee", "name_last")))
        .column(Column::from(("employee", "title")))
        .value(
            Function::from(aggregate_to_string(Column::from(("role", "display"))))
[...]

PR #123, opened 2020-04-28
Aliasing expressions

pimeys:
Helping out in https://github.com/prisma/quaint/issues/89#issuecomment-620216977

### DIFF
--- a/src/connector/sqlite.rs
+++ b/src/connector/sqlite.rs
@@ -255,6 +255,26 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_aliased_value() {
+        let conn = Sqlite::new("db/test.db").unwrap();
+        let query = Select::default().value(val!(1).alias("test"));
+        let rows = conn.select(query).await.unwrap();
+        let row = rows.get(0).unwrap();
+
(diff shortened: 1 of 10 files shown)


src/connector/sqlite.rs
```diff
@@ -254,6 +254,26 @@ mod tests {
         assert!(res.is_empty());
     }
 
+    #[tokio::test]
```
scull7, line 257: This is great!  Thank you for adding this.  In your the `quaint::ast::Select::value` documentation you use the following example:

```sql
    SELECT 
		`cats`.*,
		(`dogs`.`age` - ?)
	FROM `cats`
	INNER JOIN `dogs`
		ON `dogs`.`slave_id` = `cats`.`master_id`",
```

Would it be possible to use this to alias the `(dogs.age - ?)` expression?

pimeys: Yeah if it's an expression it holds the `Aliasable` trait now.
```rust
value((col!("dogs", "age") - val!(4)).alias("foo"))
```
Btw. I don't really like using these macros, but I also think I'm really reaching the edges how much I can bend the Rust type system here... :D

PR #125, opened 2020-04-30
Fix a broken dependency with sqlite feature

cryslith:
Commit 6a898ee3b3081a3803da18297d23021f21ee78dc introduced a dependency on `tokio/sync` for the sqlite feature, but didn't add it to the Cargo.toml file.  Commit ac264e7de92d0315c46f94e56d06b33b6f5ab0be attempted to fix this but only fixed it for the `single-sqlite` feature, leaving `full-sqlite` broken.  This PR fixes both cases by having the `sqlite` feature depend directly on `tokio/sync`.


pimeys:
Thank you. I'll set our tests to test all possible combinations tomorrow so we'll catch these better. Sorry for slow response, been quite stressful weekend trying to travel back home from a different country :)

PR #126, opened 2020-04-30
sqlite: open the database as main

cryslith:
This allows referring to possibly-nonexistent tables with their unqualified names in raw commands, and is more consistent with the behavior of the Postgres and MySQL backends.

In particular, it enables creating tables using the `Queryable` interface.  For instance, one can write `db.cmd_raw("CREATE TABLE my_table (x INTEGER);")` to create a table, instead of `db.cmd_raw("CREATE TABLE db_name.my_table (x INTEGER);")` (where `db_name` is the value of the `db_name` parameter in the database URL).  This is useful because without it, one would have to re-parse the original database URL in order to get the value of `db_name`, and then thread that value through all code that needs to create a table.

For backwards compatibility, the database file is still attached as `db_name` too, so it's mapped as both `main` and `db_name`.

### DIFF
--- a/src/connector/sqlite.rs
+++ b/src/connector/sqlite.rs
@@ -111,15 +111,15 @@ impl TryFrom<&str> for Sqlite {
 
     fn try_from(path: &str) -> crate::Result<Self> {
         let params = SqliteParams::try_from(path)?;
+        let file_path = params.file_path;
 
-        let conn = rusqlite::Connection::open_in_memory()?;
+        let conn = rusqlite::Connection::open(file_path.as_str())?;
 
         if let Some(timeout) = params.socket_timeout {
             conn.busy_timeout(timeout)?;
         };
 
         let client = Mutex::new(conn);
-        let file_path = params.file_path;
 
         Ok(Sqlite { client, file_path })
     }


tomhoule:
We have a fairly large codebase relying on quaint at https://github.com/prisma/prisma-engines

I think the rationale for the change is solid. We could test prisma with this change and merge if we don't find any issue, wdyt @pimeys ?

pimeys:
Yeah. We basically always have the database indicator in all our queries. `Model` is `User` and the `Schema` is `Blog`, so all the queries will be `User.Blog`, and `user.db` is attached as `User`.

Might be a trickier change.

tomhoule:
Sorry for neglecting this PR for so long. So I did try this patch on [the prisma repo](https://github.com/prisma/prisma-engines), and it seems like the queries on the `ATTACH`ed database are now read-only, which is a problem since we do add the attached name everywhere. I think the way forward would be for us to no longer do that, and we can merge this PR. What do you think @pimeys ?

pimeys:
Yes.

cryslith:
Closing since #192 was merged

PR #127, opened 2020-05-06
Rename "conjuctive" to "conjunctive".

cryslith:
Since this is an API-breaking change, it might be nice to get it in before 2.0.0.


tomhoule:
I think this is fairly uncontroversial, so we should merge this, wdyt @pimeys ?

pimeys:
:D Sure. Thanks

PR #128, opened 2020-05-06
Add #[must-use] to DBIO

cryslith:
This produces a warning if the user accidentally drops the DBIO without awaiting it.


tomhoule:
Thanks for the PR! I think this is a net improvement, so I'm going to go ahead and merge, feel free to protest if you see a problem I didn't @pimeys  :)

PR #129, opened 2020-05-07
Set mobc max_idle to the same as max_open by default …

tomhoule:
...and make it configurable.

max_idle is basically a less flexible alternative to max_idle_lifetime and is easier to misuse.

The default setting in mobc (2 idle connections maximum) was causing massive churn in the connection pool with 3 threads in our tests.


pimeys:
I'd introduce this as a method to the `Builder` instead, allowing everybody to tune the setting.

PR #131, opened 2020-05-07
Microsoft SQL Server Support

pimeys:
Single query support for the SQL Server. For MARS support, we might want a new trait that allows multiple queries in one statement.


src/ast/insert.rs
```diff
@@ -3,40 +3,82 @@ use crate::ast::*;
...
+    /// let (sql, _) = Sqlite::build(query.on_conflict(OnConflict::DoNothing))?;
     ///
     /// assert_eq!("INSERT OR IGNORE INTO `users` DEFAULT VALUES", sql);
+    /// # Ok(())
+    /// # }
     /// ```
+    ///
+    /// With Microsoft SQL server not supporting `IGNORE` in the `INSERT`
+    /// statement, the `INSERT` is converted to a `MERGE` statement. For it to work
+    /// in a correct way, the table should know all unique indices of the actual table.
```
tomhoule, line 47: nitpick: it would be easier to understand by specifying a bit more, like "the quaint table should know all unique indices of the actual database table".

src/ast/row.rs
```diff
@@ -38,6 +38,15 @@ impl<'a> Row<'a> {
...
 }
 
+impl<'a> IntoIterator for Row<'a> {
+    type Item = Expression<'a>;
+    type IntoIter = std::vec::IntoIter<Self::Item>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.values.into_iter()
+    }
+}
```
tomhoule, line 48: :+1:

src/connector/connection_info.rs
```diff
@@ -243,6 +277,14 @@ impl SqlFamily {
             _ => false,
         }
     }
+
+    #[cfg(feature = "mssql")]
+    pub fn is_mssql(&self) -> bool {
+        match self {
+            SqlFamily::Mssql => true,
+            _ => false,
+        }
```
tomhoule, line 286: I've started to use `matches!(self, SqlFamily::Mssql)` for these, but the match is definitely readable too.

src/connector/mssql/error.rs
```diff
@@ -0,0 +1,125 @@
...
+                    .message()
+                    .split('.')
+                    .nth(1)
+                    .unwrap()
+                    .split(' ')
+                    .last()
+                    .unwrap()
+                    .split("'")
+                    .nth(1)
+                    .unwrap();
```
tomhoule, line 57: woah

src/connector/mysql/conversion.rs
```diff
@@ -68,31 +68,52 @@ impl TakeRow for my::Row {
...
+                    MYSQL_TYPE_TIMESTAMP
+                    | MYSQL_TYPE_TIME
+                    | MYSQL_TYPE_DATE
+                    | MYSQL_TYPE_DATETIME
+                    | MYSQL_TYPE_YEAR
+                    | MYSQL_TYPE_NEWDATE
+                    | MYSQL_TYPE_TIMESTAMP2
+                    | MYSQL_TYPE_DATETIME2
+                    | MYSQL_TYPE_TIME2 => Value::DateTime(None),
+                    MYSQL_TYPE_VARCHAR | MYSQL_TYPE_VAR_STRING | MYSQL_TYPE_STRING => Value::Text(None),
```
tomhoule, line 104: I think we have to check the encoding here to determine if it's text or bytes. Same for the blob variants.

src/connector/queryable.rs
```diff
@@ -71,7 +71,7 @@ pub trait Queryable: Send + Sync {
 #[async_trait]
 pub trait TransactionCapable: Queryable
 where
-    Self: Sized + Sync,
+    Self: Sized,
```
tomhoule, line 74: :+1:

src/pooled.rs
```diff
@@ -75,6 +76,34 @@
...
+//! JDBC standard. It's quite similar to the others: the parameters are delimited
+//! with a `;` instead of `?` or `&`, and the parameter names are camelCase instead
+//! of snake_case.
+//!
+//! - `encrypt` if set to `true` encrypts all traffic over TLS. If `false`, only
+//!   the login details are encrypted.
+//! - `user` sets the login name.
+//! - `password` sets the login password.
+//! - `database` sets the database to connect to.
+//! - `trustServerCertificate` if set to `true`, accepts any kind of certificate
```
tomhoule, line 91: super nitpicky: it would be good to document the default for this one imo

src/ast/column.rs
```diff
@@ -1,35 +1,70 @@
...
+    }
+
+    /// Sets the default value for the column.
+    pub fn default<V>(mut self, value: V) -> Self
+    where
+        V: Into<Value<'a>>,
+    {
+        self.default = Some(value.into());
+        self
+    }
```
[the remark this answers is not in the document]
pimeys: In cases when you insert ignore into some table, and you hit unique constraints, you might not have all possible unique fields defined in your insert. In these cases we must get the default value to join with the target table.

src/connector/mysql/conversion.rs
pimeys: Can you be a bit more elaborate? What more should we check here?

tomhoule: It's the `column.character_set()` thing I hit in the type mapping work. This line on master: https://github.com/prisma/quaint/blob/c2b2b4af698e2440d81e4fd2820a787ab42ca56a/src/connector/mysql/conversion.rs#L51

pimeys: Would this be enough? https://github.com/prisma/quaint/pull/131/commits/35d635dfaa565571f8e9afa35e83748c071892bd

PR #133, opened 2020-05-20
Pgbouncer mode for PostgreSQL

pimeys:
Introducing a new flag for PostgreSQL: `pgbouncer`.

When `true`, after `BEGIN` of each transaction a query of `DEALLOCATE ALL` is executed to clean existing prepared statements from the session to prevent collisions.

To be able to use Quaint with pgbouncer's transaction mode, every query needs to be inside a transaction and the connection string should have `pgbouncer=true` set.

### DIFF
--- a/src/connector/postgres.rs
+++ b/src/connector/postgres.rs
@@ -3,7 +3,7 @@ mod error;
 
 use crate::{
     ast::{Query, Value},
-    connector::{metrics, queryable::*, ResultSet, DBIO},
+    connector::{metrics, queryable::*, ResultSet, Transaction, DBIO},
     error::{Error, ErrorKind},
     visitor::{self, Visitor},
 };
@@ -44,6 +44,7 @@ impl std::fmt::Debug for PostgresClient {
 #[derive(Debug)]
 pub struct PostgreSql {
     client: PostgresClient,
+    pg_bouncer: bool,
     socket_timeout: Option<Duration>,
 }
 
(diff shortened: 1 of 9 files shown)


src/connector/postgres.rs
```diff
@@ -223,9 +225,15 @@ impl PostgresUrl {
...
         let mut connect_timeout = None;
+        let mut pg_bouncer = false;
 
         for (k, v) in url.query_pairs() {
             match k.as_ref() {
+                "pgbouncer" => {
+                    pg_bouncer = v
+                        .parse()
+                        .map_err(|_| Error::builder(ErrorKind::InvalidConnectionArguments).build())?;
+                }
```
janpio, line 236: Does this check for `true`/`false` or just presence of `pgbouncer´? (Me no Rust...)

pimeys: true/false

pimeys: so `pg_bouncer` variable is a boolean, `parse` will parse whatever type is on the left side from the string.

src/pooled.rs
```diff
@@ -51,6 +51,12 @@
 //! - `connect_timeout` defined in seconds (default: 5). Connecting to a
 //!   database will return a `ConnectTimeout` error if taking more than the
 //!   defined value.
+//! - `pgbouncer` either `true` or `false`. If set, allows usage with the
+//!   pgBouncer connection pool in transaction mode. Additionally a transaction
+//!   is required for every query for the mode to work. When starting a new
+//!   transaction, a deallocation query `DEALLOCATE ALL` is executed right after
+//!   `BEGIN` to avoid possible collisions with statements created in other
+//!   sessions.
```
janpio, line 59: +1

PR #142, opened 2020-06-11
Use `map` instead of `map_and_drop` on mysql query

pimeys:
The `map_and_drop` in `mysql_async` will freeze forever if using it with
stored procedure calls, such as `CALL StoredProcedure()`. For some
reason `map` works.


pimeys:
Fixes https://github.com/prisma/prisma/issues/2679

PR #143, opened 2020-06-17
Enabling statement cache for PostgreSQL

pimeys:
Adds a new parameter `statement_cache_size` to enable statement caching for PostgreSQL. By default 0, meaning no cache. If `pgbouncer` flag is on, cache will be disabled.

issue: https://github.com/prisma/prisma/issues/2720

### DIFF
--- a/src/connector/postgres.rs
+++ b/src/connector/postgres.rs
@@ -9,6 +9,7 @@ use crate::{
 };
 use async_trait::async_trait;
 use futures::{future::FutureExt, lock::Mutex};
+use lru_cache::LruCache;
 use native_tls::{Certificate, Identity, TlsConnector};
 use percent_encoding::percent_decode;
 use postgres_native_tls::MakeTlsConnector;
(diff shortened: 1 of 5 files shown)


src/connector/postgres.rs
```diff
@@ -226,6 +236,7 @@ impl PostgresUrl {
         let mut socket_timeout = None;
         let mut connect_timeout = None;
         let mut pg_bouncer = false;
+        let mut statement_cache_size = 0;
```
tomhoule, line 239: We'll probably figure things out as we start using it, but for other quaint users, if it works we may want to default to a non-zero size.

pimeys: I don't want to enable this to everybody by default. Should be opt-in.

tomhoule: There's no other way to reuse prepared statements currently, so this should at least to be well documented.

pimeys: Ok. Making the default here to 500. Let's see...

PR #156, opened 2020-07-01
Lints

yoshuawuyts:
o/ This PR is an attempt to ensure that `quaint` will continue to compile in future versions of the compiler and introduce a few new code quality measures that will help automate review.

This PR currently fails CI because it turns out that we, in fact, are not future compatible. In particular the issue we're running into is https://github.com/rust-lang/rust/issues/62411, which will likely be turned into an error for the 2021 edition. But beyond that it seems there are a fair number of other issues we may want to address.

Anyway, I found myself having a spare 30 mins today and was curious how this would turn out. I'm opening this as a draft PR to share what's going on, and if this is deemed useful perhaps someone could go through and address the errors and warnings this has surfaced. Thanks!

### DIFF
--- a/.buildkite/docker.sh
+++ b/.buildkite/docker.sh
@@ -24,6 +24,7 @@ docker run -w /build --network test-net -v $BUILDKITE_BUILD_CHECKOUT_PATH:/build
     -e TEST_MYSQL=mysql://prisma:prisma@test-mysql:3306/prisma \
     -e TEST_PSQL=postgres://prisma:prisma@test-postgres:5432/prisma \
     -e TEST_MSSQL="sqlserver://test-mssql:1433;user=SA;password=$MSSQL_SA_PASSWORD;trustServerCertificate=true" \
+    -e RUSTFLAGS="-Dwarnings" \
     prismagraphql/build:test cargo test --features full,json-1,uuid-0_8,chrono-0_4,tracing-log,serde-support
 
 exit_code=$?
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -103,6 +103,11 @@
 //! # Ok(())
 //! # }
 //! ```
+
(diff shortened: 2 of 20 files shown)


.buildkite/docker.sh
```diff
@@ -24,6 +24,7 @@ docker run -w /build --network test-net -v $BUILDKITE_BUILD_CHECKOUT_PATH:/build
     -e TEST_MYSQL=mysql://prisma:prisma@test-mysql:3306/prisma \
     -e TEST_PSQL=postgres://prisma:prisma@test-postgres:5432/prisma \
     -e TEST_MSSQL="sqlserver://test-mssql:1433;user=SA;password=$MSSQL_SA_PASSWORD;trustServerCertificate=true" \
+    -e RUSTFLAGS="-Dwarnings" \
```
yoshuawuyts, line 27: This ensures we don't ship warnings to our main branch.

src/lib.rs
```diff
@@ -103,6 +103,11 @@
 //! # Ok(())
 //! # }
 //! ```
+
+#![forbid(unsafe_code, future_incompatible, rust_2018_idioms)]
+#![deny(missing_debug_implementations, nonstandard_style)]
+#![warn(missing_docs, unreachable_pub)]
+
```
yoshuawuyts, line 110: This is the set of lints introduced. This includes linting for `rust_2018_idioms` which are things that are disallowed in the current edition and can prevent us from using rust 2018 in the future. And `future_incompatible` which are things that will be disallowed in a future edition (due next year).

yoshuawuyts:
Talked with @pimeys and some of the current issues we're running into will be resolved with the switch away from `tokio-postgres` to `sqlx`. We can pick this patch back up after that!

PR #160, opened 2020-07-22
Add a hack to fix a pgBouncer bug with enums

pimeys:
Now tokio-postgres needs a special branch to fix panics when using
pgBouncer and enums.

And this escalates us of branching out rust_decimal due to it depending
on `ToSql` and `FromSql` from postgres.

And finally we must branch tiberius due to it depending on rust_decimal,
and the git version not being compatible with the release version.

Sigh.


pimeys:
Fixes https://github.com/prisma/prisma/issues/2921

PR #163, opened 2020-07-28
Fix a quite demented param handling in sqlite

pimeys:
- We don't need supported/unsupported division in params
- And we should store the full url to the pool's manager
- Otherwise parameters get filtered out

Fixes https://github.com/prisma/prisma/issues/2955

### DIFF
--- a/src/connector/sqlite.rs
+++ b/src/connector/sqlite.rs
@@ -22,6 +22,7 @@ pub struct Sqlite {
     pub(crate) file_path: String,
 }
 
+#[derive(Debug)]
 pub struct SqliteParams {
     pub connection_limit: Option<usize>,
     /// This is not a `PathBuf` because we need to `ATTACH` the database to the path, and this can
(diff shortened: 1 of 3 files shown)


src/connector/sqlite.rs
```diff
@@ -50,25 +49,17 @@ impl TryFrom<&str> for SqliteParams {
...
-                        (String::from(splitted[0]), String::from(splitted[1]))
-                    })
-                    .collect::<Vec<(String, String)>>()
-                    .into_iter()
-                    .partition(|(k, _)| official.contains(&k.as_str()));
-
-                for (k, v) in unsupported.into_iter() {
+                let params = path_parts.last().unwrap().split('&').map(|kv| {
+                    let splitted: Vec<&str> = kv.split('=').collect();
+                    (String::from(splitted[0]), String::from(splitted[1]))
```
tomhoule, line 59: nitpick: do these need to be allocated, or would matching on string slices below be enough, and only allocating what we need?

pimeys: Done like donuts.

PR #166, opened 2020-07-29
Implement possibility to use multiple table-likes in FROM.

dpetrick:
- Adds `.from` to select.
- Changes `table` in `Select` to `tables` and makes it a vector of tables instead of an option of a single table.

### DIFF
--- a/src/ast/select.rs
+++ b/src/ast/select.rs
@@ -112,7 +112,7 @@ impl<'a> Select<'a> {
     /// # use quaint::{ast::*, visitor::{Visitor, Sqlite}};
     /// # fn main() -> Result<(), quaint::error::Error> {
     /// let query = Select::from_table("users")
-    ///     .from(Table::from(Select::default().value(1)).alias("num"))
+    ///     .and_from(Table::from(Select::default().value(1)).alias("num"))
     ///     .column(("users", "name"))
     ///     .value(Table::from("num").asterisk());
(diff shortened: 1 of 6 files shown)


src/ast/select.rs
```diff
@@ -101,11 +101,35 @@ impl<'a> Select<'a> {
...
+    ///     .column(("users", "name"))
+    ///     .value(Table::from("num").asterisk());
+    ///
+    /// let (sql, _) = Sqlite::build(query)?;
+    ///
+    /// assert_eq!("SELECT `users`.`name`, `num`.* FROM `users`, (SELECT ?) AS `num`", sql);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn from<T>(mut self, table: T) -> Self
```
pimeys, line 125: Name this `and_from`, `from` is reserved for `From` trait.

dpetrick: Done

pimeys: Actually, now when I think about this, can this be anything else than another `Select`? If so, please change the signature into `T: Into<Select>`...

pimeys: And if this always needs to have an alias, then this would be:

```rust
pub fn and_from(mut self, select: impl Into<Select>, alias: impl Into<Cow<'a, str>>) -> Self {
    ..
}
```

PR #170, opened 2020-08-14
Add raw comparators

dpetrick:
Allows to inject unsupported comparators as raw values: `col.compare_raw("ILIKE", "foo%")`

### DIFF
--- a/src/ast/compare.rs
+++ b/src/ast/compare.rs
@@ -452,7 +452,7 @@ pub trait Comparable<'a> {
         T: Into<Expression<'a>>,
         V: Into<Expression<'a>>;
 
-    /// Tests if the value is not between two given values.
+    /// Compares two expressions with a custom operator.
     ///
     /// ```rust
     /// # use quaint::{ast::*, visitor::{Visitor, Sqlite}};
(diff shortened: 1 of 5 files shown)


src/visitor.rs
```diff
@@ -744,6 +744,11 @@ pub trait Visitor<'a> {
                 self.write(" AND ")?;
                 self.visit_expression(*right)
             }
+            Compare::Raw(left, comp, right) => {
+                self.visit_expression(*left)?;
+                self.write(format!(" {} ", comp))?;
```
tomhoule, line 749: nitpick: we can avoid the string allocation here:

```rust
 self.visit_expression(*left)?;
 self.write(" ")?;
self.write(comp)?;
self.write(" ")?;
self.visit_expression(*right)
```

src/ast/compare.rs
```diff
@@ -448,6 +451,27 @@ pub trait Comparable<'a> {
     where
         T: Into<Expression<'a>>,
         V: Into<Expression<'a>>;
+
+    /// Tests if the value is not between two given values.
```
tomhoule, line 455: Nitpick: I think this is copy-pasted. Maybe something like `compares two expressions with a custom operator` or something like that.

PR #171, opened 2020-08-27
Log if the query was successful (or not)

pimeys:
Needs to have `LOG_QUERIES` set to `1` to be visible.

With tracing, adds a new key `result`, with a possible value `success` or `error` to the query log.

Without tracing, changes the rendering to either:

```
query: "SELECT 1", params: [] (success in 0ms)
```

or:

```
query: "SELECT aseas", params: [] (error in 0ms)
```

Closes https://github.com/prisma/prisma-engines/issues/879

### DIFF
--- a/src/connector/metrics.rs
+++ b/src/connector/metrics.rs
@@ -16,12 +16,18 @@ where
     let end = Instant::now();
 
     if *crate::LOG_QUERIES {
+        let result = match res {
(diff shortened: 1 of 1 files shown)


tomhoule (approved):
I didn't grasp the whole conversation, but this looks like a nice improvement to me!

PR #176, opened 2020-09-10
Fix of precision when storing/loading floats

pimeys:
The rust_decimal has changed their floating point conversion that has
tendency to lose precision in some cases. We now take the approach of
going from Decimal -> String -> f64, which is slower but keeps the precision.

### DIFF
--- a/src/connector/postgres/conversion.rs
+++ b/src/connector/postgres/conversion.rs
@@ -464,7 +464,7 @@ impl<'a> ToSql for Value<'a> {
                 f.to_sql(ty, out)
             }),
             (Value::Real(decimal), &PostgresType::FLOAT8) => decimal.map(|decimal| {
-                let f = decimal.to_f64().expect("decimal to f64 conversion");
+                let f = decimal.to_string().parse::<f64>().expect("decimal to f64 conversion");
                 f.to_sql(ty, out)
             }),
             (Value::Array(decimals), &PostgresType::FLOAT4_ARRAY) => decimals.as_ref().map(|decimals| {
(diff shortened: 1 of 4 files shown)


src/connector/postgres/conversion.rs
```diff
@@ -477,7 +477,10 @@ impl<'a> ToSql for Value<'a> {
             (Value::Array(decimals), &PostgresType::FLOAT8_ARRAY) => decimals.as_ref().map(|decimals| {
                 let f: Vec<f64> = decimals
                     .into_iter()
-                    .filter_map(|v| v.as_decimal().and_then(|decimal| decimal.to_f64()))
+                    .filter_map(|v| {
+                        v.as_decimal()
+                            .and_then(|decimal| decimal.to_string().parse::<f64>().ok())
+                    })
```
tomhoule, line 483: In this case, it looks like we are doing &[f64] -> &[Decimal] -> &[String] -> Vec<f64> (it's not the case, but schematically). Would it be possible to not convert to decimal instead? v.as_f64() directly, maybe?

tomhoule (approved):
It's not the first problem we have with rust_decimal :/

src/connector/postgres/conversion.rs
pimeys: Wait, no, this goes from `Value::Real(Decimal)` -> `Decimal` -> `String` -> `f64`...

tomhoule: ah indeed, if it's already a decimal we have to do that

PR #177, opened 2020-09-10
A better error for invalid MySQL dates.

pimeys:
This addresses the case of old MySQL feature of storing `0000-00-00 00:00:00` for invalid dates. This cannot be converted to chrono, thus it's better to error gracefully instead of triggering a panic.


tomhoule (approved):
Nice! I thought we already returned a user-facing error for that instead of a panic. Much better.

pimeys:
What panics here is `chrono`, it doesn't return an error but just panics. We can't have that...

PR #182, opened 2020-09-15
Add extra conversions

tomhoule:
- Postgres BYTEA
- SQLite datetime text format
- deserialize_bytes in serde module

### DIFF
--- a/src/connector/postgres/conversion.rs
+++ b/src/connector/postgres/conversion.rs
@@ -123,6 +123,15 @@ impl GetRow for PostgresRow {
                     }
                     None => Value::Bytes(None),
                 },
+                #[cfg(feature = "array")]
+                PostgresType::BYTEA_ARRAY => match row.try_get(i)? {
+                    Some(val) => {
+                        let val: Vec<Vec<u8>> = val;
+                        let byteas = val.into_iter().map(Value::bytes);
+                        Value::array(byteas)
+                    }
+                    None => Value::Array(None),
+                },
                 PostgresType::MONEY => match row.try_get(i)? {
                     Some(val) => {
                         let val: NaiveMoney = val;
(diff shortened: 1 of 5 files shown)


tomhoule:
I think this (sqlite and postgres bytea) should be tested, but we (@do4gr and me) don't have a lot of bandwidth at the moment. I'll try to find time.

tomhoule:
I'll probably drop the BYTEA work (we don't need it immediately anymore, and too much hurry to spend time on testing it) and just keep the SQLite datetime FromSql implementation

pimeys:
Tests are commented out, just put them back and see if you get them green. https://github.com/prisma/quaint/blob/master/src/tests/types/postgres.rs#L365

tomhoule:
@pimeys the postgres bytea test. Can't figure out how to test the sqlite conversion in the roundtrip tests. It's necessary for the `serde` feature to work with datetimes on SQLite.

src/connector/postgres/conversion.rs
```diff
@@ -116,6 +116,13 @@ impl GetRow for PostgresRow {
                     }
                     None => Value::Real(None),
                 },
+                PostgresType::BYTEA => match row.try_get(i)? {
+                    Some(val) => {
+                        let val: &[u8] = val;
```
yoshuawuyts, line 121: Is this line needed? If it's for inference of `try_get`, maybe use turbofish there instead?

pimeys: yeah it's the `try_get` above: https://docs.rs/tokio-postgres/0.5.5/tokio_postgres/row/struct.Row.html#method.try_get

tomhoule: It's that way to align with the other branches. I think the tokio-postgres doc show something like that as well. Agreed I'd prefer a turbofish.

PR #185, opened 2020-09-17
Use tokio's block in place to not block the executor for sqlite queries

pimeys:
Let's see if this makes benchmarks any faster with SQLite!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,7 +91,7 @@ tracing-core = { version = "0.1", optional = true }
 
 mobc = { version = "0.5.7", optional = true }
 bytes = { version = "0.5", optional = true }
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
From [the Tokio docs](https://docs.rs/tokio/0.2.22/tokio/task/fn.block_in_place.html):

> [...] any other code running concurrently in the same task will be suspended during the call to block_in_place. This can happen e.g. when using the join! macro. To avoid this issue, use spawn_blocking instead.

The way I'm reading this it means that calls to `Future::timeout` will not work. Given this is a library and we don't control all uses this seems like a problem waiting to happen. We should use `tokio::task::spawn_blocking` instead.

PR #186, opened 2020-09-17
Test and fix raw_cmd() on MySQL

tomhoule:
closes https://github.com/prisma/quaint/issues/187

### DIFF
--- a/src/connector/mysql.rs
+++ b/src/connector/mysql.rs
@@ -354,8 +354,8 @@ impl Queryable for Mysql {
 #[cfg(test)]
 mod tests {
     use super::MysqlUrl;
+    use crate::tests::test_api::mysql::CONN_STR;
     use crate::{connector::Queryable, error::*, single::Quaint};
-    use crate::{prelude::Select, tests::test_api::mysql::CONN_STR};
     use url::Url;
 
     #[test]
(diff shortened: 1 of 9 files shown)


src/connector/mysql.rs
```diff
@@ -391,4 +391,52 @@ mod tests {
...
+        let conn = Quaint::new(&CONN_STR).await.unwrap();
+
+        conn.raw_cmd("DROP DATABASE `raw_cmd_test`;").await.unwrap();
+        conn.raw_cmd("CREATE DATABASE `raw_cmd_test`;").await.unwrap();
+
+        let conn = Quaint::new(&CONN_STR.replace("dbname", "raw_cmd_test")).await.unwrap();
+
+        conn.raw_cmd(
+            "
+            CREATE TABLE `raw_cmd_test`.`testtable` (id INTEGER PRIMARY KEY);
```
pimeys, line 419: temporary table?

tomhoule: This is crashing. Let's discuss this later, but I'm getting pessimistic about fixing this.

src/connector/mysql.rs
```diff
@@ -185,27 +185,24 @@ impl MysqlUrl {
...
 
     pub(crate) fn to_opts_builder(&self) -> my::OptsBuilder {
-        let mut config = my::OptsBuilder::new();
-
-        config.user(Some(self.username()));
-        config.pass(self.password());
-        config.db_name(Some(self.dbname()));
+        let mut config = my::OptsBuilder::default().user(Some(self.username()))
+        .pass(self.password())
+        .db_name(Some(self.dbname()));
```
tomhoule, line 190: note: run `cargo fmt` manually. I think it doesn't work automatically on my setup because of the feature flags.

src/connector/mysql.rs
```diff
@@ -391,4 +401,60 @@ mod tests {
         let err = res.unwrap_err();
         assert!(matches!(err.kind(), ErrorKind::AuthenticationFailed { user } if user == "WRONG"));
     }
+
+    #[tokio::test]
```
pimeys, line 405: This could also be `test_all_connector` in `query.rs`.

pimeys (changes requested):
Maybe extend the tests to all databases, code is good.

src/tests/query/error.rs
```diff
@@ -271,3 +272,65 @@ async fn garbage_datetime_values(api: &mut dyn TestApi) -> crate::Result<()> {
...
+    }
+
+    Ok(())
+}
+
+#[test_each_connector(tags("mysql"))]
+async fn should_execute_multi_statement_queries_with_raw_cmd(api: &mut dyn TestApi) -> crate::Result<()> {
+    let conn = api.conn();
+
+    conn.raw_cmd(
```
pimeys, line 300: Maybe use temporary tables here?

src/tests/query/error.rs
```diff
@@ -271,3 +271,65 @@ async fn garbage_datetime_values(api: &mut dyn TestApi) -> crate::Result<()> {
...
+    if api.conn().connection_info().sql_family().is_mysql() {
+        let error_message = result.unwrap_err().to_string();
+        assert_eq!(error_message, "Error accessing result set, column not found: YOLO");
+    }
+
+    Ok(())
+}
+
+#[test_each_connector(tags("mysql"))]
+async fn should_execute_multi_statement_queries_with_raw_cmd(api: &mut dyn TestApi) -> crate::Result<()> {
```
tomhoule, line 296: This test is challenging to make cross-database, the create table abstraction in `TestApi` doesn't really work because we specifically want to test multiple statements in the same script. So it would have a bunch of database-specific code, or a new create table abstraction.

pimeys: Something similar to this? https://github.com/prisma/quaint/blob/master/src/tests/test_api/mysql.rs#L43-L56

tomhoule: It wouldn't test what we want to test, because we want the 4 statements in one call.

pimeys: We could have an additional function to that...

tomhoule:
@pimeys So let's maybe wait until next release is done to merge this.

tomhoule:
@pimeys, from our last discussions I'm going to assume it's ok to merge this now, so we can get the fix down the pipeline and test it as soon as possible.

pimeys:
Gogogo.

PR #189, opened 2020-09-22
Allow setting the schema for MSSQL

pimeys:
- Defaults to `dbo`
- Added to every query as an identifier

### DIFF
--- a/src/connector/mssql.rs
+++ b/src/connector/mssql.rs
@@ -29,6 +29,7 @@ pub(crate) struct MssqlQueryParams {
     user: Option<String>,
     password: Option<String>,
     database: String,
+    schema: String,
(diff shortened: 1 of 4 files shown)


src/connector/mssql.rs
```diff
@@ -286,6 +295,7 @@ impl MssqlUrl {
                 let user = params.remove("user");
                 let password = params.remove("password");
                 let database = params.remove("database").unwrap_or_else(|| String::from("master"));
+                let schema = params.remove("schema").unwrap_or_else(|| String::from("dbo"));
```
pimeys, line 298: The connection-string crate will be of good use here...

tomhoule (approved):

PR #190, opened 2020-09-23
Allow setting the SQL Server isolation level

pimeys:
Helps our tests to use `SNAPSHOT` isolation, preventing deadlocks.

### DIFF
--- a/src/connector/mssql.rs
+++ b/src/connector/mssql.rs
@@ -9,7 +9,14 @@ use crate::{
 };
 use async_trait::async_trait;
 use futures::lock::Mutex;
-use std::{collections::HashMap, convert::TryFrom, fmt::Write, future::Future, time::Duration};
+use std::{
+    collections::HashMap,
+    convert::TryFrom,
+    fmt::{self, Write},
+    future::Future,
+    str::FromStr,
+    time::Duration,
+};
 use tiberius::*;
 use tokio::{net::TcpStream, time::timeout};
 use tokio_util::compat::{Compat, Tokio02AsyncWriteCompatExt};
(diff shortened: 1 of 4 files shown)


yoshuawuyts:
@pimeys can you say more on why you're choosing `SNAPSHOT` isolation over `SERIALIZABLE`?

pimeys:
`SNAPSHOT` isolation uses MVCC locking instead of the default and old 2-phase locking of `READ COMMITTED`. All our open source databases have been using MVCC since forever, but due to backwards compatibility reasons the default of SQL Server is still `READ COMMITED`.

This means when anybody writes things to the database, as in our case in the tests to the schema, the reads are locked and if we're unlucky, we get a deadlock and need to retry transactions.

With MVCC locking there's a separate version table that gives much more fine-grained locking, preventing deadlocks and giving fast concurrency. The downside of this then is how the tempdb of the server can grow, due to it needing to store the version information of the row before collected.

src/connector/mssql.rs
```diff
@@ -34,6 +34,7 @@ pub(crate) struct MssqlQueryParams {
     connection_limit: Option<usize>,
     socket_timeout: Option<Duration>,
     connect_timeout: Option<Duration>,
+    transaction_isolation_level: Option<String>,
```
yoshuawuyts, line 37: Given isolation levels are enumerable, could we make this take an `Option<IsolationLevel>` instead?

yoshuawuyts (changes requested):

src/pooled.rs
```diff
@@ -98,8 +98,13 @@
 //! - `connectTimeout` defined in seconds (default: 5). Connecting to a
 //!   database will return a `ConnectTimeout` error if taking more than the
 //!   defined value.
-//! - `connection:imit` defines the maximum number of connections opened to the
+//! - `connectionLimit` defines the maximum number of connections opened to the
 //!   database.
+//! - `schema` the name of the lookup schema. Only stored to the connection,
+//!   must be used in every query to be effective.
+//! - `isolationLevel` the transaction isolation level. Possible values:
+//!   `READ UNCOMMITTED`, `READ COMMITED`, `REPETABLE READ`, `SNAPSHOT`,
```
tomhoule, line 106: nitpick: I think there and in the next file there is a type on `READ COMMITTED` and on `REPEATABLE READ`.

pimeys:
yeah fixed that already.

Sent from [ProtonMail](https://protonmail.ch), encrypted email based in Switzerland.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

src/pooled.rs
pimeys: yeah fixed that already.

pimeys:
Fixed all.

pimeys:
And why not `SERIALIZABLE`

> Range locks are placed in the range of key values that match the search conditions of each statement executed in a transaction. This blocks other transactions from updating or inserting any rows that would qualify for any of the statements executed by the current transaction. This means that if any of the statements in a transaction are executed a second time, they will read the same set of rows. The range locks are held until the transaction completes. This is the most restrictive of the isolation levels because it locks entire ranges of keys and holds the locks until the transaction completes. Because concurrency is lower, use this option only when necessary. This option has the same effect as setting HOLDLOCK on all tables in all SELECT statements in a transaction.

We don't need this much of isolation for tests. We want as little locks as possible due to the tests running in parallel.

PR #192, opened 2020-09-30
sqlite: open files using the main database

tomhoule:
This allows referring to tables with their unqualified names in raw commands.

This is just a rebased version of @cryslith PR: https://github.com/prisma/quaint/pull/126


tomhoule:
discussed with @pimeys and this is fine to merge if no downstream breakage happens — let's revert it if it causes issues.

PR #194, opened 2020-10-01
Xml support

dpetrick:
Adds `Value::Xml` and handling of that type. The implementation is simplistic at the moment, e.g. equals comparisons for a XML field are done as string operations.

### DIFF
--- a/src/connector/postgres/conversion.rs
+++ b/src/connector/postgres/conversion.rs
@@ -457,7 +457,7 @@ impl GetRow for PostgresRow {
 
 impl ToColumnNames for PostgresStatement {
     fn to_column_names(&self) -> Vec<String> {
-        self.columns().into_iter().map(|c| c.name().into()).collect()
+        self.columns().iter().map(|c| c.name().into()).collect()
     }
 }
 
@@ -517,7 +517,7 @@ impl<'a> ToSql for Value<'a> {
             #[cfg(feature = "uuid-0_8")]
             (Value::Array(values), &PostgresType::UUID_ARRAY) => values.as_ref().map(|values| {
                 let parsed_uuid: Vec<Uuid> = values
-                    .into_iter()
+                    .iter()
                     .filter_map(|v| v.to_string().and_then(|v| v.parse().ok()))
                     .collect();
                 parsed_uuid.to_sql(ty, out)
@@ -531,7 +531,7 @@ impl<'a> ToSql for Value<'a> {
             (Value::Array(values), &PostgresType::INET_ARRAY) | (Value::Array(values), &PostgresType::CIDR_ARRAY) => {
                 values.as_ref().map(|values| {
                     let parsed_ip_addr: Vec<std::net::IpAddr> = values
-                        .into_iter()
+                        .iter()
                         .filter_map(|v| v.to_string().and_then(|s| s.parse().ok()))
                         .collect();
                     parsed_ip_addr.to_sql(ty, out)
@@ -551,7 +551,7 @@ impl<'a> ToSql for Value<'a> {
             (Value::Array(values), &PostgresType::BIT_ARRAY) | (Value::Array(values), &PostgresType::VARBIT_ARRAY) => {
                 values.as_ref().map(|values| {
                     let bitvecs: Vec<BitVec> = values
-                        .into_iter()
+                        .iter()
                         .filter_map(|val| val.as_str().map(|s| string_to_bits(s)))
                         .collect::<crate::Result<Vec<_>>>()?;
 
--- a/src/connector/postgres/conversion.rs
+++ b/src/connector/postgres/conversion.rs
@@ -404,6 +404,22 @@ impl GetRow for PostgresRow {
                     }
                     None => Value::Array(None),
                 },
+                #[cfg(feature = "xml")]
+                PostgresType::XML => match row.try_get(i)? {
+                    Some(val) => {
+                        let val: String = val;
+                        Value::xml(val)
+                    }
+                    None => Value::Xml(None),
+                },
+                #[cfg(feature = "array")]
+                PostgresType::XML_ARRAY => match row.try_get(i)? {
+                    Some(val) => {
+                        let val: Vec<String> = val;
+                        Value::array(val)
+                    }
+                    None => Value::Array(None),
+                },
                 ref x => match x.kind() {
                     Kind::Enum(_) => match row.try_get(i)? {
                         Some(val) => {
--- a/src/connector/postgres/conversion.rs
+++ b/src/connector/postgres/conversion.rs
@@ -569,7 +569,7 @@ impl<'a> ToSql for Value<'a> {
             (Value::Array(vec), _) => vec.as_ref().map(|vec| vec.to_sql(ty, out)),
             #[cfg(feature = "json-1")]
             (Value::Json(value), _) => value.as_ref().map(|value| value.to_sql(ty, out)),
-            #[cfg(feature = "json-1")]
+            #[cfg(feature = "xml")]
             (Value::Xml(value), _) => value.as_ref().map(|value| value.to_sql(ty, out)),
             #[cfg(feature = "uuid-0_8")]
             (Value::Uuid(value), _) => value.map(|value| value.to_sql(ty, out)),
--- a/src/connector/postgres/conversion.rs
+++ b/src/connector/postgres/conversion.rs
@@ -25,6 +25,18 @@ pub fn conv_params<'a>(params: &'a [Value<'a>]) -> Vec<&'a (dyn types::ToSql + S
     params.iter().map(|x| x as &(dyn ToSql + Sync)).collect::<Vec<_>>()
 }
 
+struct XmlString(pub String);
+
+impl<'a> FromSql<'a> for XmlString {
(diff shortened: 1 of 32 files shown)


src/connector/postgres/conversion.rs
```diff
@@ -569,6 +569,8 @@ impl<'a> ToSql for Value<'a> {
             (Value::Array(vec), _) => vec.as_ref().map(|vec| vec.to_sql(ty, out)),
             #[cfg(feature = "json-1")]
             (Value::Json(value), _) => value.as_ref().map(|value| value.to_sql(ty, out)),
+            #[cfg(feature = "json-1")]
```
pimeys, line 572: ?

pimeys (changes requested):

src/connector/postgres/conversion.rs
dpetrick: Copy pasta, sorry!

src/tests/types/postgres.rs
```diff
@@ -276,6 +276,17 @@ test_type!(jsonb_array(
...
+test_type!(xml(postgres, "xml", Value::Xml(None), Value::xml("<test>1</test>",)));
+
+#[cfg(all(feature = "xml", feature = "array"))]
+test_type!(xml_array(
+    postgres,
+    "xml[]",
+    Value::Array(None),
+    Value::array(vec!["<test>1</test>", "<test>2</test>",])
+));
+
```
pimeys, line 289: Similar tests for SQL Server and MySQL?

src/visitor/mssql.rs
```diff
@@ -79,31 +79,65 @@ impl<'a> Visitor<'a> for Mssql<'a> {
...
                 self.write(" = ")?;
-                self.visit_expression(Expression::from(right_kind))?;
+
+                match self {
+                    #[cfg(feature = "xml")]
+                    _ if left_xml => {
+                        self.surround_with("CAST(", " AS NVARCHAR(MAX))", |x| x.visit_expression(right))?
+                    }
+                    _ => self.visit_expression(right)?,
+                };
```
pimeys, line 107: What about testing these (and other supported databases) in https://github.com/prisma/quaint/blob/master/src/tests/query.rs?

src/visitor/mssql.rs
```diff
@@ -79,31 +79,65 @@ impl<'a> Visitor<'a> for Mssql<'a> {
     }
 
     fn visit_equals(&mut self, left: Expression<'a>, right: Expression<'a>) -> visitor::Result {
```
pimeys, line 81: This and the `visit_not_equals` can be written without cloning (which can be very expensive here):

```patch
diff --git a/src/ast/expression.rs b/src/ast/expression.rs
index c349a91..0a2b2cf 100644
--- a/src/ast/expression.rs
+++ b/src/ast/expression.rs
@@ -51,11 +51,7 @@ impl<'a> Expression<'a> {
 
     #[cfg(feature = "xml")]
     pub(crate) fn is_xml_value(&self) -> bool {
-        match &self.kind {
-            ExpressionKind::Parameterized(Value::Xml(_)) => true,
-            ExpressionKind::Value(expr) => expr.is_xml_value(),
-            _ => false,
-        }
+        self.kind.is_xml_value()
     }
 
     pub(crate) fn is_asterisk(&self) -> bool {
@@ -186,6 +182,17 @@ pub enum ExpressionKind<'a> {
     Value(Box<Expression<'a>>),
 }
 
+impl<'a> ExpressionKind<'a> {
+    #[cfg(feature = "xml")]
+    pub(crate) fn is_xml_value(&self) -> bool {
+        match self {
+            Self::Parameterized(Value::Xml(_)) => true,
+            Self::Value(expr) => expr.is_xml_value(),
+            _ => false,
+        }
+    }
+}
+
 /// A quick alias to create an asterisk to a table.
 pub fn asterisk() -> Expression<'static> {
     Expression {
diff --git a/src/visitor/mssql.rs b/src/visitor/mssql.rs
index a5b47d7..15f7b57 100644
--- a/src/visitor/mssql.rs
+++ b/src/visitor/mssql.rs
@@ -1,4 +1,5 @@
 use super::Visitor;
+use crate::prelude::Aliasable;
 use crate::prelude::Query;
 use crate::{
     ast::{Column, Expression, ExpressionKind, Insert, IntoRaw, Merge, OnConflict, Order, Ordering, Row, Values},
@@ -79,18 +80,32 @@ impl<'a> Visitor<'a> for Mssql<'a> {
     }
 
     fn visit_equals(&mut self, left: Expression<'a>, right: Expression<'a>) -> visitor::Result {
-        match (&left.kind, &right.kind) {
+        match (left.kind, right.kind) {
             // we can't compare with tuples, so we'll convert it to an AND
             (ExpressionKind::Row(left), ExpressionKind::Row(right)) => {
-                self.visit_multiple_tuple_comparison(left.clone(), Values::from(iter::once(right.clone())), false)?;
+                self.visit_multiple_tuple_comparison(left, Values::from(iter::once(right)), false)?;
             }
-            (_left_kind, _right_kind) => {
+            (left_kind, right_kind) => {
+                let (l_alias, r_alias) = (left.alias, right.alias);
+
                 #[cfg(feature = "xml")]
-                let left_xml = left.is_xml_value();
+                let (left_xml, right_xml) = (left_kind.is_xml_value(), right_kind.is_xml_value());
+
+                let mut left = Expression::from(left_kind);
+
+                if let Some(alias) = l_alias {
+                    left = left.alias(alias);
+                }
+
+                let mut right = Expression::from(right_kind);
+
+                if let Some(alias) = r_alias {
+                    right = right.alias(alias);
+                }
 
                 match self {
                     #[cfg(feature = "xml")]
-                    _ if right.is_xml_value() => {
+                    _ if right_xml => {
                         self.surround_with("CAST(", " AS NVARCHAR(MAX))", |x| x.visit_expression(left))?
                     }
                     _ => self.visit_expression(left)?,
@@ -112,18 +127,32 @@ impl<'a> Visitor<'a> for Mssql<'a> {
     }
 
     fn visit_not_equals(&mut self, left: Expression<'a>, right: Expression<'a>) -> visitor::Result {
-        match (&left.kind, &right.kind) {
+        match (left.kind, right.kind) {
             // we can't compare with tuples, so we'll convert it to an AND
             (ExpressionKind::Row(left), ExpressionKind::Row(right)) => {
-                self.visit_multiple_tuple_comparison(left.clone(), Values::from(iter::once(right.clone())), true)?;
+                self.visit_multiple_tuple_comparison(left, Values::from(iter::once(right)), true)?;
             }
-            (_left_kind, _right_kind) => {
+            (left_kind, right_kind) => {
+                let (l_alias, r_alias) = (left.alias, right.alias);
+
                 #[cfg(feature = "xml")]
-                let left_xml = left.is_xml_value();
+                let (left_xml, right_xml) = (left_kind.is_xml_value(), right_kind.is_xml_value());
+
+                let mut left = Expression::from(left_kind);
+
+                if let Some(alias) = l_alias {
+                    left = left.alias(alias);
+                }
+
+                let mut right = Expression::from(right_kind);
+
+                if let Some(alias) = r_alias {
+                    right = right.alias(alias);
+                }
 
                 match self {
                     #[cfg(feature = "xml")]
-                    _ if right.is_xml_value() => {
+                    _ if right_xml => {
                         self.surround_with("CAST(", " AS NVARCHAR(MAX))", |x| x.visit_expression(left))?
                     }
                     _ => self.visit_expression(left)?,
```
```

**Review by pimeys — CHANGES_REQUESTED** [review]

**Comment by pimeys** [comment]
Naturally squash this before rebasing!

File: src/tests/types/postgres.rs (hunk 2 above)
Comment (dpetrick) [reply src/tests/types/postgres.rs:289]: SQL server has those test. SQLite and MySQL do not have XML types and are out of scope.

File: src/visitor/mssql.rs (hunk 3 above)
Comment (dpetrick) [reply src/visitor/mssql.rs:107]: Good point, didn't know about these.

Comment (pimeys) [reply src/visitor/mssql.rs:107]: They're a bit hidden, but a quite easy way to do proper integration testing.

File: src/visitor/mssql.rs (hunk 4 above)
Comment (dpetrick) [reply src/visitor/mssql.rs:81]: Was wondering if there's an easy way to get rid of the clone calls.

Comment (pimeys) [reply src/visitor/mssql.rs:81]: This is not that "easy", but something to put you going. You can make this even better if you like, kind of noisy my changes...

Comment (pimeys) [reply src/visitor/mssql.rs:81]: The aliasing could be nicer, but better go to sleep and stop working goddammit...

File: src/connector/postgres/conversion.rs [hunk 5]
```diff
@@ -404,6 +416,22 @@ impl GetRow for PostgresRow {
...
                 },
+                #[cfg(feature = "xml")]
+                PostgresType::XML => match row.try_get(i)? {
+                    Some(val) => {
+                        let val: XmlString = val;
+                        Value::xml(val.0)
+                    }
+                    None => Value::Xml(None),
+                },
+                #[cfg(feature = "array")]
```
pimeys, line 427: `#[cfg(all(feature = "array", feature = "xml"))]`

pimeys (changes requested):
Super nitpick, otherwise good to go.

src/connector/postgres/conversion.rs
dpetrick: Good catch.

PR #198, opened 2020-10-08
Add API for in-memory SQLite

tomhoule:


src/connector/connection_info.rs
```diff
@@ -119,6 +121,8 @@ impl ConnectionInfo {
             ConnectionInfo::Mssql(url) => url.schema(),
             #[cfg(feature = "sqlite")]
             ConnectionInfo::Sqlite { db_name, .. } => db_name,
+            #[cfg(feature = "sqlite")]
+            ConnectionInfo::InMemorySqlite => panic!("called schema_name() on InMemorySqlite"),
```
tomhoule, line 125: not too happy about this. Wdyt?

tomhoule:
There is intentionally no way to express in-memory databases as URLs. This would need to be discussed and thought through first.

tomhoule:
TODO: docs for `Sqlite::new_in_memory()` and `single::Quaint::new_in_memory()`.

src/connector/connection_info.rs
pimeys: Why not `Option<&str>`?

src/single.rs
```diff
@@ -160,6 +160,14 @@ impl Quaint {
         Ok(Self { inner, connection_info })
     }
 
+    #[cfg(feature = "sqlite")]
+    pub fn new_in_memory() -> crate::Result<Quaint> {
```
pimeys, line 164: I was first saying hell no, but now when I think about this, actually it makes sense. The only problem is you can't pass connection params through this interface.

src/connector/connection_info.rs
tomhoule: We use it a lot in the other variants, it would be painful to handle nulls there. But it's definitely possible.

src/single.rs
tomhoule: What params would you pass?

src/connector/connection_info.rs
pimeys: I don't want a library to panic. At least a `Result`.

tomhoule: I think there's another way: still keep the attached name, but attach with `:memory:` (so no file). That way we can keep this returning a string.

src/single.rs
tomhoule: Note: if we go with my last suggestion on the other discussion, we should pass the attached name here.

tomhoule: (optionally)

pimeys:
Few tests fail on Windows. You could maybe fix them with your changes here?

```
test src\lib.rs - (line 32) ... FAILED
test src\lib.rs - (line 51) ... FAILED

failures:

---- src\lib.rs - (line 32) stdout ----
Test executable failed (exit code 1).

stderr:
Error: Error { kind: QueryError(SqliteFailure(Error { code: CannotOpen, extended_code: 14 }, Some("unable to open database file: ///tmp/example.db"))), original_code: Some("14"), original_message: Some("unable to open database file: ///tmp/example.db") }


---- src\lib.rs - (line 51) stdout ----
Test executable failed (exit code 1).

stderr:
Error: Error { kind: QueryError(SqliteFailure(Error { code: CannotOpen, extended_code: 14 }, Some("unable to open database file: ///tmp/example.db"))), original_code: Some("14"), original_message: Some("unable to open database file: ///tmp/example.db") }



failures:
    src\lib.rs - (line 32)
    src\lib.rs - (line 51)

test result: FAILED. 78 passed; 2 failed; 1 ignored; 0 measured; 0 filtered out
```

tomhoule:
Hmm, so these paths would need to be valid on windows. We can probably get valid windows paths from `tempfile`.

pimeys (approved):

PR #202, opened 2020-10-12
Use connection-string

yoshuawuyts:
Uses `connection-string` for JDBC parsing. This required mutable access to `properties`, so I've pushed `connection-string@0.1.5` which provides access to `properties_mut`.

Also I've made it so the JDBC parser handles errors in the string rather than silently continuing. This should make it easier to debug malformed strings (instead of a param not showing up, we'll return an error if we found one).

`connection-string` doesn't know how to _encode_ anything yet, so I haven't touched the ado.net logic. Hope this is good!

### DIFF
--- a/.buildkite/docker.sh
+++ b/.buildkite/docker.sh
@@ -30,7 +30,7 @@ docker run -w /build --network test-net -v $BUILDKITE_BUILD_CHECKOUT_PATH:/build
     -e TEST_MYSQL=mysql://prisma:prisma@test-mysql:3306/prisma \
     -e TEST_MYSQL8=mysql://prisma:prisma@test-mysql-8:3306/prisma \
     -e TEST_PSQL=postgres://prisma:prisma@test-postgres:5432/prisma \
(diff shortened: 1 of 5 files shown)


yoshuawuyts:
Oh, so in order for tests to pass I've had to rewrite the MSSQL connection string to always require the `jdbc:sqlserver` prefix. I'm assuming we weren't doing this because Prisma doesn't want to expose `jdbc:` to end-users. I get why we wouldn't want to expose that through Prisma; but should we be rewriting connection strings in quaint? I would assume the answer is "probably not", but keen to hear more.

pimeys (dismissed):

pimeys (changes requested):
We should edit the docs in `single.rs` and `pooled.rs` to reflect the changes to the required connection string format. E.g. the added `jdbc:` prefix!

pimeys:
Closing in favor of https://github.com/prisma/quaint/pull/215/

PR #203, opened 2020-10-13
Log successful acquisition of db connection

dpetrick:


src/pooled/manager.rs
```diff
@@ -105,7 +105,14 @@ impl Manager for QuaintManager {
                 use crate::connector::Mssql;
                 Ok(Box::new(Mssql::new(url.clone()).await?) as Self::Connection)
             }
+        };
+
+        #[cfg(feature = "tracing-log")]
```
pimeys, line 110: You could also do `if cfg!(feature = "tracing-log")` { .. }` here.

src/pooled/manager.rs
```diff
@@ -105,7 +105,14 @@ impl Manager for QuaintManager {
...
+        {
+            conn.iter()
+                .for_each(|_| tracing::info!("Acquired database connection."));
```
pimeys, line 113: Maybe debug is better level, or do we want to spam the info layer?

pimeys (changes requested):

PR #209, opened 2020-10-20
To cut a dependency hell for now, we'll copy some code

pimeys:
THIS IS WHAT IT ALL MEANS PEOPLE!

Copied the `ToSql`/`FromSql` code from rust-decimal to cut a bad dependency hell and get many of our deps back to crates.io. Some of the code is cleaned using the knowledge from SQLx, but it's still quite a lot :(

### DIFF
--- a/src/connector/postgres/conversion/decimal.rs
+++ b/src/connector/postgres/conversion/decimal.rs
@@ -1,0 +1,293 @@
+//! We implement the postgres conversions here to fix a disastrous dependency hell.
+
+use byteorder::{BigEndian, ReadBytesExt};
+use bytes::{BufMut, BytesMut};
+use postgres_types::{to_sql_checked, FromSql, IsNull, ToSql, Type};
+use rust_decimal::{prelude::Zero, Decimal};
+use std::convert::TryInto;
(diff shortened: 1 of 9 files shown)


src/connector/postgres/conversion/decimal.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    fn accepts(ty: &Type) -> bool {
+        match ty {
+            &Type::NUMERIC => true,
+            _ => false,
+        }
+    }
+
+    to_sql_checked!();
+}
```
tomhoule, line 293: Are there tests we could also copy?

tomhoule (approved):
ok ok ok

src/connector/postgres/conversion/decimal.rs
pimeys: I don't know do I want those tests...

pimeys: Need to write my own.

pimeys: I'll do that before merging.

PR #218, opened 2020-10-30
Replace rust-decimal with bigdecimal

pimeys:
Adds also support for Float and Double without using BigDecimal.


Cargo.toml
```diff
@@ -65,14 +68,13 @@ metrics = "0.12"
...
 percent-encoding = "2"
-rust_decimal = "1.8"
 thiserror = "1.0"
 url = "2.1"
 
 either = { version = "1.6", optional = true }
 byteorder = { default-features = false, optional = true, version = "1.3" }
 base64 = { version = "0.11.0", optional = true }
-chrono = { version = "0.4", optional = true }
+chrono_ = { version = "0.4", optional = true, package = "chrono" }
```
tomhoule, line 77: Is it a trick to decouple the `chrono` feature from the `chrono` optional crate?

pimeys: Ah, kinda, this was because earlier the `chrono` feature enabled `tiberius/chrono` too, but then I noticed that enabling `chrono` would always download tiberius, which is not what I want. So it really seems we cannot optionally enable `chrono` for tiberius without fetching the package always.

Which is a bummer.

src/ast/values.rs
```diff
@@ -107,7 +108,8 @@ impl<'a> fmt::Display for Value<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
tomhoule, line 108: It's weird to me we have a Display impl for Value. I don't remember where we use it, is there a specific purpose to it?

src/ast/values.rs
```diff
@@ -152,7 +156,18 @@ impl<'a> From<Value<'a>> for serde_json::Value {
     fn from(pv: Value<'a>) -> Self {
         let res = match pv {
             Value::Integer(i) => i.map(|i| serde_json::Value::Number(Number::from(i))),
-            Value::Real(d) => d.map(|d| serde_json::to_value(d).unwrap()),
+            Value::Float(f) => f.map(|f| match Number::from_f64(f as f64) {
+                Some(number) => serde_json::Value::Number(number),
```
tomhoule, line 160: Is this really needed? I would imagine serde_json can deal with special float values

src/ast/values.rs
```diff
@@ -395,23 +427,51 @@ impl<'a> Value<'a> {
         }
     }
 
-    /// `true` if the `Value` is a real value.
-    pub fn is_real(&self) -> bool {
-        matches!(self, Value::Real(_))
+    /// Returns a f64 if the value is a numeric value and the underlying decimal
+    /// can be converted, otherwise `None`.
```
tomhoule, line 431: I think this docstring is not accurate anymore

src/ast/values.rs
```diff
@@ -395,23 +427,51 @@ impl<'a> Value<'a> {
...
+            Value::Double(Some(f)) => Some(*f as f64),
+            _ => None,
+        }
     }
 
-    /// Returns a f64 if the value is a real value and the underlying decimal can be converted, otherwise `None`.
-    pub fn as_f64(&self) -> Option<f64> {
+    /// Returns a f32 if the value is a numeric value and the underlying decimal
+    /// can be converted, otherwise `None`.
+    pub fn as_f32(&self) -> Option<f32> {
```
tomhoule, line 441: same as previous comment

src/ast/values.rs
```diff
@@ -395,23 +427,51 @@ impl<'a> Value<'a> {
...
+        match self {
+            Value::Float(Some(f)) => Some(*f),
+            _ => None,
+        }
+    }
+
+    /// `true` if the `Value` is a numeric value or can be converted to a one.
```
tomhoule, line 448: typo: "to a one"

tomhoule (approved):
lots of work, congrats!

src/ast/values.rs
pimeys: `f32` or `f64` do not implement `Display`. Maybe we could use the `{:?}` here?

tomhoule: We could take the JS behaviour as a reference there: 

```
JSON.stringify({ n: Infinity })
"{\"n\":null}"
JSON.stringify({ n: NaN })
"{\"n\":null}"
JSON.stringify({ n: 1.0 })
"{\"n\":1}"
```

src/ast/values.rs
pimeys: the logger...

PR #226, opened 2020-12-01
Fix unique criteria error in SQL Server, update Tiberius for wire fixes

pimeys:
Fixes error handling a bit. Has Tiberius fixes from https://github.com/prisma/tiberius/pull/105

### DIFF
--- a/src/connector/mssql/error.rs
+++ b/src/connector/mssql/error.rs
@@ -97,6 +97,19 @@ impl From<tiberius::error::Error> for Error {
 
                 builder.build()
             }
+            tiberius::error::Error::Server(e) if e.code() == 1505 => {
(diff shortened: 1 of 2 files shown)


src/connector/mssql/error.rs
```diff
@@ -97,6 +97,19 @@ impl From<tiberius::error::Error> for Error {
 
                 builder.build()
             }
+            tiberius::error::Error::Server(e) if e.code() == 1505 => {
+                let mut splitted = e.message().split('\'');
+                let index = splitted.nth(3).unwrap().to_string();
```
tomhoule, line 102: When we have time to tackle this, we should really think of a more robust approach for error metadata extraction :/

tomhoule (approved):

PR #227, opened 2020-12-02
Convert connection string errors correctly

pimeys:
Part of https://github.com/prisma/prisma-engines/issues/1241

### DIFF
--- a/src/connector/sqlite.rs
+++ b/src/connector/sqlite.rs
@@ -111,7 +111,10 @@ impl TryFrom<&str> for Sqlite {
         let params = SqliteParams::try_from(path)?;
         let file_path = params.file_path;
 
-        let conn = rusqlite::Connection::open(file_path.as_str())?;
+        let conn = rusqlite::Connection::open(file_path.as_str()).map_err(|e| {
+            let kind = ErrorKind::DatabaseUrlIsInvalid(e.to_string());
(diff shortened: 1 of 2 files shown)


src/connector/sqlite.rs
```diff
@@ -111,7 +111,10 @@ impl TryFrom<&str> for Sqlite {
         let params = SqliteParams::try_from(path)?;
         let file_path = params.file_path;
 
-        let conn = rusqlite::Connection::open(file_path.as_str())?;
+        let conn = rusqlite::Connection::open(file_path.as_str()).map_err(|e| {
```
tomhoule, line 114: Are we sure this is always a connection string issue? From the context, it feels like it could be for other reasons (file not readable, etc.)

pimeys: This is the only place where it throws if we have a faulty string... Maybe we should just not have this error for sqlite at all?

pimeys: (And remove this change)

tomhoule: Hmm yes I think I'd remove it, if we can't be more specific, because it's confusing if we return a url parse error for a file missing permissions. Also SQLite connection strings aren't URLs

tomhoule (approved):

PR #232, opened 2021-01-08
Skip using the internal pool of MySQL

pimeys:
We have no need for it, and we can fail immediately when creating the connection, if needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,7 @@ thiserror = "1.0"
 url = "2.1"
 
 either = { version = "1.6", optional = true }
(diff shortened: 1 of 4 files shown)


tomhoule (approved):
removing one level of indirection sounds good! The behaviour when the connections fail should be the same, right? In pooled mode we have healthchecks, in single mode it aligns with postgres

pimeys:
It aligns with Postgres and it's a part of a fix to a bug ticket I took to myself last Friday from Will https://github.com/prisma/prisma/issues/4335

PR #235, opened 2021-01-14
Add an error message if pool is saturated

pimeys:
Better error for the issue in https://github.com/prisma/prisma/issues/4280

Edit: Change of plans. Now we introduce a new parameter `pool_timeout` to error out if pool is saturated. `connect_timeout` will now only be used when connecting to the database.

Both parameters are disabled by default.

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -25,8 +25,10 @@ pub(crate) mod postgres;
 #[cfg(feature = "sqlite")]
 pub(crate) mod sqlite;
 
+#[cfg(any(feature = "mssql", feature = "postgresql", feature = "mysql"))]
 use std::time::Duration;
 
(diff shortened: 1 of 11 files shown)


dpetrick (approved):
Looking good. Maybe link to the docs for connection params in the error.
What's the reasoning behind the "close to saturation" case?

pimeys:
I mean, this can error out in two ways:

1) we have all connections in use, we get a connection timeout because we couldn't get a connection in a timely manner
2) the network is bonkers and we can't just open a connection at all.

Too bad we have no control to see if the error is from forming a connection or if the pool is maxed out. I'm making an educated guess here that after we get a timeout, if we're close to a saturation, we probably are maxed out, but having a race here and at the time of the error we might have available connections again.

tomhoule (approved):
LGTM!

PR #240, opened 2021-01-19
fix: convert tinyint(1) to bool only when val is 0 or 1

Weakky:
Fixes https://github.com/prisma/prisma/issues/4981

### DIFF
--- a/src/connector/mysql/conversion.rs
+++ b/src/connector/mysql/conversion.rs
@@ -175,8 +175,7 @@ impl TypeIdentifier for my::Column {
     }
 
     fn is_bool(&self) -> bool {
-        (self.column_type() == ColumnType::MYSQL_TYPE_TINY && self.column_length() == 1)
-            || (self.column_type() == ColumnType::MYSQL_TYPE_BIT && self.column_length() == 1)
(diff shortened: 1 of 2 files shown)


dpetrick (approved):

pimeys:
What would happen here if the user wants to use the tinyint as tinyint, and wants to store 0 or 1?

Weakky:
> What would happen here if the user wants to use the tinyint as tinyint, and wants to store 0 or 1?

The answer, for now, would be: coerce the value back to an int just like we're doing in the QE.

I don't know the exact reasons why we're converting `tinyint(1)`s to `boolean`s in the first place, so it'd be helpful to know that to properly formulate an answer.

pimeys:
Now, to the next try, I'd say better would be to go and edit

```rust
    fn is_bool(&self) -> bool {
        (self.column_type() == ColumnType::MYSQL_TYPE_TINY && self.column_length() == 1)
            || (self.column_type() == ColumnType::MYSQL_TYPE_BIT && self.column_length() == 1)
    }
```

Now, we could maybe continue converting `BIT(1)` still to a boolean. Therefore you'd change the test to use `BIT(1)`, but you'd remove the first check here and add a few more tests to see that `TINYINT` maps to an from an integer.

pimeys (approved):

PR #241, opened 2021-01-21
Skip timeouts in mobc, we handle them always in Quaint anyhow

pimeys:


janpio (approved):
This should disable the default mentioned at https://docs.rs/mobc/0.5.12/mobc/struct.Builder.html#method.get_timeout 👍

dpetrick (approved):

PR #242, opened 2021-01-27
Default value keyword support

dpetrick:
Adds support for `default` expression. Can be used in inserts for example: `INSERT INTO table (id, a, b) VALUES (1, default, "b1"), (2, "a2, "b2"),`


pimeys:
Otherwise fine, but please add an integration test to https://github.com/prisma/quaint/blob/master/src/tests/query.rs

Run it only against databases you know support this syntax.

PR #252, opened 2021-02-11
fix(sqlite): do NOT create two db connections

Weakky:
Fixes https://github.com/prisma/prisma-engines/issues/1481

This removes support for the path params "db_name". Quaint was using the `ATTACH DATABASE` statement as a hack to scope the db under some custom name.
The `ATTACH DATABASE` statement is initially designed to create another connection to a different db so that one can query multiple dbs. This was causing subtle issues with transactions.

Namely, if inside a transaction, a user would "manually" execute (with `executeRaw`) a write _without_ prefixing the tables with the custom db name (using one connection), and that the QE would also execute a read, this time _with_ the tables prefixed with the db_name (therefore with another connection), this would cause two different connections attempting to do READ/WRITES inside a transaction, causing locks as explained in this doc https://sqlite.org/isolation.html

SQLite _has_ as default db name, it's called `main`. This PR defaults the `db_name` to `main`. This enables:

- The QE to keep referring to the tables using some alias (`main`)
- The users to refer to their tables without using any aliases (that they're not aware of anyway). This is possible because when no alias is used, SQLite will default to using the `main` connection.

### DIFF
--- a/src/connector/sqlite.rs
+++ b/src/connector/sqlite.rs
@@ -14,7 +14,7 @@ use rusqlite::NO_PARAMS;
 use std::{collections::HashSet, convert::TryFrom, path::Path, time::Duration};
 use tokio::sync::Mutex;
 
-pub(crate) const DEFAULT_SQLITE_SCHEMA_NAME: &str = "quaint";
+pub(crate) const DEFAULT_SQLITE_SCHEMA_NAME: &str = "main";
 
 /// A connector interface for the SQLite database
 #[cfg_attr(feature = "docs", doc(cfg(feature = "sqlite")))]
@@ -56,7 +56,6 @@ impl TryFrom<&str> for SqliteParams {
             Err(Error::builder(ErrorKind::DatabaseUrlIsInvalid(path.to_str().unwrap().to_string())).build())
         } else {
             let mut connection_limit = None;
-            let mut db_name = None;
             let mut socket_timeout = None;
 
             if path_parts.len() > 1 {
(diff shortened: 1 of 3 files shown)


src/connector/sqlite.rs
```diff
@@ -10,18 +10,18 @@ use crate::{
...
+pub(crate) const DEFAULT_SQLITE_SCHEMA_NAME: &str = "main";
 
 /// A connector interface for the SQLite database
 #[cfg_attr(feature = "docs", doc(cfg(feature = "sqlite")))]
 pub struct Sqlite {
     pub(crate) client: Mutex<rusqlite::Connection>,
     /// This is not a `PathBuf` because we need to `ATTACH` the database to the path, and this can
     /// only be done with UTF-8 paths. This is `None` for purely in-memory databases.
+    #[allow(dead_code)]
     pub(crate) file_path: Option<String>,
```
Weakky, line 25: Not sure if it's useful to keep that field around as it's a public struct that could be used in existing projects

tomhoule: `pub(crate)` means it's not exposed outside the crate, so if you get a warning it's not used anywhere, for sure.

pimeys (approved):
Great, did you test this in Prisma engines already?

When merging, remember to squash.

src/connector/sqlite.rs
pimeys: And you should not expose fields from a public crate. If you want to give an access to the data, you'd use a method instead, allowing us to change their way of working without breaking any public apis.

Weakky: Good, I'll get rid of it before merging

Weakky:
> Great, did you test this in Prisma engines already?
> 
> When merging, remember to squash.

I did, it works. It actually works without even this fix since the QE fix no longer passes the `db_name` params anyway, so the attach db stuff is no longer called. PR's here, I just need to add a simple test https://github.com/prisma/prisma-engines/pull/1644

Issue #259, opened 2021-02-25
Support `ROW_TO_JSON`


GrandChaman:
Hi !

It would be a nice addition to be able to return rows as `JSON` in `Postgres`.

pimeys:
We already have that feature!

Did you check the [documentation](https://prisma.github.io/quaint/quaint/serde/index.html)?

pimeys:
`serde_json::Value` implements `Deserialize`, so you can use the serde functions with that type if you don't want to have a struct.

pimeys:
Nevermind, I was too quick...

pimeys:
Too long day already. Merged.

PR #260, opened 2021-02-25
Added support for 'ROW_TO_JSON' function for postgresql

GrandChaman:
PR to support [the issue](https://github.com/prisma/quaint/issues/259) I just opened.

Maybe it would be nice to create a new `Function` type that would be accessible only for the `Postgresql` visitor. In its current state, every visitor could use the newly exposed function, but only `Postgresql` would be compatible.

I'd love some feedback !

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Changelog
+## Unreleased
+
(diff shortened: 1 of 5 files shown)


pimeys:
No this is great. Thank you!

PR #265, opened 2021-03-03
Instrument method calls

pimeys:
We get span data and can count the time taken per function.

Breaking: we'll now force tracing for everybody. You can probably still use the crate without, but might miss some logs.

### DIFF
--- a/src/connector/mssql.rs
+++ b/src/connector/mssql.rs
@@ -237,6 +237,7 @@ pub struct Mssql {
 
 impl Mssql {
     /// Creates a new connection to SQL Server.
+    #[tracing::instrument(name = "new_connection", skip(url))]
     pub async fn new(url: MssqlUrl) -> crate::Result<Self> {
         let config = Config::from_jdbc_string(&url.connection_string)?;
         let tcp = TcpStream::connect_named(&config).await?;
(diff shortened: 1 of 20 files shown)


src/connector/mssql.rs
```diff
@@ -237,6 +237,7 @@ pub struct Mssql {
 
 impl Mssql {
     /// Creates a new connection to SQL Server.
+    #[tracing::instrument(name = "new_connection", skip(url))]
```
tomhoule, line 240: I haven't used the `name` argument before, good to know it exists

src/connector/mssql.rs
```diff
@@ -276,16 +277,19 @@ impl Mssql {
...
         self.query_raw(&sql, &params[..]).await
     }
 
+    #[tracing::instrument(skip(self, q))]
     async fn execute(&self, q: Query<'_>) -> crate::Result<u64> {
         let (sql, params) = visitor::Mssql::build(q)?;
         self.execute_raw(&sql, &params[..]).await
     }
 
+    #[tracing::instrument(skip(self))]
```
tomhoule, line 292: I'm concerned sql and params could be large or contain information we don't want to leak, here — maybe worrying too much?

src/connector/mssql.rs
```diff
@@ -326,6 +330,7 @@ impl Queryable for Mssql {
         .await
     }
 
+    #[tracing::instrument(skip(self))]
```
tomhoule, line 333: same as previous comment

tomhoule (approved):
awesome! I'm conflicted on the feature removal, because tracing is clearly better and it's saving complexity in quaint to remove all these conditionals, but it's also a larger dependency. I still think it's the right thing to remove. :+1:

src/connector/mssql.rs
pimeys: I want them to the tracing views. It's important to see the SQL there.

src/connector/mssql.rs
pimeys: And same answer too. I try to skip as much params as I can from instrumentation, but the SQL query is something we must have there for our users in the future. Think of this whole thing as a product in the Prisma Studio :)

src/connector/mssql.rs
pimeys: Maybe we can skip the params?

PR #267, opened 2021-03-07
Expose maximum connection lifetimes

Sytten:
This adds the `max_lifetime` on the builder and exposes `max_connection_lifetime` and `max_idle_connection_lifetime` via the URL parser.
Needed for serverless environments where it is not a good idea to never close connections, see https://github.com/prisma/prisma/issues/5977

I did try to respect the coding style of each parser but there is a refactoring job to be done down there to make them uniform, it's a bit of a mess.

### DIFF
--- a/src/connector/mssql.rs
+++ b/src/connector/mssql.rs
@@ -75,6 +75,8 @@ pub(crate) struct MssqlQueryParams {
     connect_timeout: Option<Duration>,
     pool_timeout: Option<Duration>,
     transaction_isolation_level: Option<IsolationLevel>,
+    max_connection_lifetime: Option<Duration>,
+    max_idle_connection_lifetime: Option<Duration>,
 }
 
 #[derive(Debug, Clone, Copy)]
@@ -186,6 +188,16 @@ impl MssqlUrl {
     pub fn connection_string(&self) -> &str {
         &self.connection_string
     }
+
+    /// The maximum connection lifetime
+    pub fn max_connection_lifetime(&self) -> Option<Duration> {
+        self.query_params.max_connection_lifetime()
+    }
+
+    /// The maximum idle connection lifetime
+    pub fn max_idle_connection_lifetime(&self) -> Option<Duration> {
+        self.query_params.max_idle_connection_lifetime()
+    }
 }
 
 impl MssqlQueryParams {
@@ -223,6 +235,14 @@ impl MssqlQueryParams {
 
     fn pool_timeout(&self) -> Option<Duration> {
         self.pool_timeout
+    }
+
+    fn max_connection_lifetime(&self) -> Option<Duration> {
+        self.max_connection_lifetime
+    }
+
+    fn max_idle_connection_lifetime(&self) -> Option<Duration> {
+        self.max_idle_connection_lifetime
     }
 }
 
@@ -467,6 +487,16 @@ impl MssqlUrl {
             .transpose()?
             .unwrap_or(false);
 
+        let max_connection_lifetime = props
+            .remove("max_connection_lifetime")
+            .map(|param| param.parse().map(Duration::from_secs))
+            .transpose()?;
+
+        let max_idle_connection_lifetime = props
+            .remove("max_idle_connection_lifetime")
+            .map(|param| param.parse().map(Duration::from_secs))
+            .transpose()?;
+
         Ok(MssqlQueryParams {
             encrypt,
             port,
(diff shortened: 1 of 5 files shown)


Sytten:
Let me know if you want tests for that parsing somewhere

pimeys:
These are overwritten in https://github.com/prisma/prisma-engines/blob/master/query-engine/connectors/sql-query-connector/src/database/mssql.rs#L34-L35 et.al.

Sytten:
Yes will need to change that too.

janpio (approved):
Exposing these two options from our connection pool to potentially being able to use them from the Engines looks useful.

Sytten:
@pimeys @janpio I added some getters on the builder since I found this was the most straight forward way to not override the parameter in the engines. The url is parsed multiple times so it is quite confusing really (connection info in the engine, connection info in the builder and new method of the builder), but the builder getter wont lie.

pimeys:
Had a discussion with @Sytten and we decided that the defaults we set in the engines could be defined (and documented) in Quaint instead, removing the need for the getters.

src/connector/postgres.rs
```diff
@@ -254,6 +264,8 @@ impl PostgresUrl {
         let mut pool_timeout = Some(Duration::from_secs(10));
         let mut pg_bouncer = false;
         let mut statement_cache_size = 500;
+        let mut max_connection_lifetime = None;
+        let mut max_idle_connection_lifetime = None;
```
pimeys, line 268:
```suggestion
        let mut max_idle_connection_lifetime = Some(Duration::from_secs(300));
```

See: https://github.com/prisma/prisma-engines/blob/master/query-engine/connectors/sql-query-connector/src/database/postgresql.rs

src/connector/sqlite.rs
```diff
@@ -53,6 +55,8 @@ impl TryFrom<&str> for SqliteParams {
         } else {
             let mut connection_limit = None;
             let mut socket_timeout = None;
+            let mut max_connection_lifetime = None;
+            let mut max_idle_connection_lifetime = None;
```
pimeys, line 59: I think this is a good default for SQLite. See in https://github.com/prisma/prisma-engines/blob/master/query-engine/connectors/sql-query-connector/src/database/sqlite.rs#L51 we have a different value, but I'd remove it completely and let the default be infinite (SQLite doesn't hold a network connection anyhow).

src/connector/mysql.rs
```diff
@@ -121,6 +131,8 @@ impl MysqlUrl {
         let mut socket_timeout = None;
         let mut connect_timeout = Some(Duration::from_secs(5));
         let mut pool_timeout = Some(Duration::from_secs(10));
+        let mut max_connection_lifetime = None;
+        let mut max_idle_connection_lifetime = None;
```
pimeys, line 135:
```suggestion
        let mut max_idle_connection_lifetime = Some(Duration::from_secs(300));
```

And remove this in engines https://github.com/prisma/prisma-engines/blob/master/query-engine/connectors/sql-query-connector/src/database/mysql.rs

src/connector/mssql.rs
```diff
@@ -467,6 +487,16 @@ impl MssqlUrl {
             .transpose()?
             .unwrap_or(false);
 
+        let max_connection_lifetime = props
+            .remove("max_connection_lifetime")
+            .map(|param| param.parse().map(Duration::from_secs))
+            .transpose()?;
+
+        let max_idle_connection_lifetime = props
```
pimeys, line 495: Remember to remove this from the engines: https://github.com/prisma/prisma-engines/blob/master/query-engine/connectors/sql-query-connector/src/database/mssql.rs#L34

pimeys (changes requested):
Now you can try this out together with the engines by editing the topmost `Cargo.toml` in there and by adding

```toml
[patch."https://github.com/prisma/quaint"]
quaint = { path = "/path/to/local/quaint" } # this needs to be checked out to the correct branch
```

When you're happy how it works, we should merge them quite soon together due to otherwise somebody else might update quaint and get the changes beforehand, causing havoc in the worst case.

src/connector/mssql.rs
```diff
@@ -467,6 +487,16 @@ impl MssqlUrl {
...
+            .remove("max_idle_connection_lifetime")
+            .map(|param| param.parse().map(Duration::from_secs))
+            .transpose()?;
```
pimeys, line 498:
```suggestion
            .transpose()?
            .unwrap_or_else(|| Some(Duration::from_secs(300)));
```

pimeys (approved):

PR #269, opened 2021-03-12
Allow snake_case connection params for SQL Server

pimeys:
Addresses lots of confusion, e.g. issue https://github.com/prisma/prisma/issues/5897


tomhoule (approved):
Any docs that we should update?

pimeys:
Actually this just makes us compliant with the docs...

PR #272, opened 2021-03-23
fix: bump mobc version

Sytten:
This will add the FIFO queue in the connection requests and will solve a bug when requests are canceled.
Did extensive testing on that one.


pimeys (approved):
yup

PR #275, opened 2021-03-26
Accurate conversion from bigdecimal to f32 and f64

pimeys:
Quick and dirty answer to https://github.com/prisma/prisma/issues/6189


tomhoule (approved):
:'(

PR #281, opened 2021-04-06
Fix SQLite datetime conversion

dpetrick:
Rustqlite stores incoming dates like `1968-11-02T02:25:15.160Z` as millis (`-36711284840`), which fails the replaced conversion logic with all sorts of invalid datetime panics.

Proposed fix is to simply use the UTC timestamp conversion chrono offers, without additional computations in between.


pimeys:
Neat and nice.

pimeys (approved):

PR #288, opened 2021-04-20
Better typing on OUTPUT for SQL Server

pimeys:
When using the `OUTPUT` on SQL Server `INSERT`, with triggers, the trick with return list should be typed correctly. Originally we just assumed the types close enough, but of course miscalculated certain exceptions. E.g. `decimal` was always `decimal(32,16)`, which would not fit `decimal(26,0)`...

Here we extend the type hints to the column with optional parameters.

Part of https://github.com/prisma/prisma/issues/5950

### DIFF
--- a/src/tests/query.rs
+++ b/src/tests/query.rs
@@ -628,6 +628,120 @@ async fn returning_insert(api: &mut dyn TestApi) -> crate::Result<()> {
     Ok(())
 }
 
+#[cfg(all(feature = "mssql", feature = "bigdecimal"))]
+#[test_each_connector(tags("mssql"))]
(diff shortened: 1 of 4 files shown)


src/tests/query.rs
```diff
@@ -628,6 +628,120 @@ async fn returning_insert(api: &mut dyn TestApi) -> crate::Result<()> {
...
+        .await?;
+
+    assert_eq!(1, res.len());
+
+    let row = res.get(0).unwrap();
+    assert_eq!(Some(2), row["id"].as_i64());
+    assert_eq!(Some(&dec), row["val"].as_numeric());
+
+    Ok(())
+}
```
tomhoule, line 655: very nice!

tomhoule (commented):
Very nice!

PR #290, opened 2021-05-03
SQLite update + support INSERT .. RETURNING

pimeys:
The AST should Just Work (tm) now and of course raw queries can have the `RETURNING` keyword as well. The SQLite update has some datetime fixes that might break things / need to be checked.

Part of: https://github.com/prisma/prisma/issues/6876


tomhoule (approved):
:+1:

PR #292, opened 2021-05-11
Prevent decimal/float conversions where unstable. Allow casting of expressions.

pimeys:
- JSON serialization from decimal should be a string
- Do not allow writing to a decimal field with floats, or a float field with decimals
- API to cast expressions.

Part of: https://github.com/prisma/prisma/issues/6902

### DIFF
--- a/src/ast/castable.rs
+++ b/src/ast/castable.rs
@@ -238,18 +238,22 @@ impl<'a> CastType<'a> {
         self
     }
 
+    #[cfg(feature = "postgresql")]
     pub(crate) fn postgres_enabled(&self) -> bool {
         self.on_databases.contains(CastDatabase::Postgres)
     }
 
+    #[cfg(feature = "mysql")]
     pub(crate) fn mysql_enabled(&self) -> bool {
         self.on_databases.contains(CastDatabase::Mysql)
     }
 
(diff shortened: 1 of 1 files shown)


tomhoule (approved):

sumeetattree:
@pimeys This looks great! Just what I have been looking for to add query support to enum types. Anything I can do to help fast track this PR?

sumeetattree:
@pimeys Slight correction in postgres visitor (Not sure about other databases), the custom type name should be double quoted.

```rust
let column = Column::from(col_name).cast_as(CastType::custom("UserKind"));
```
The query output is:
```sql
SELECT    
  "public"."users".*    
FROM    
  "public"."users"    
WHERE    
  ("kind") :: UserKind = $1 # I know crazy type name, but sometimes external databases have stuff like this
```

"UserKind" will actually be converted to "userkind" by postgres. To avoid this the type must be quoted. `"\"UserKind\""`.  The `CastType::Custom` should automatically quote the typenames.

pimeys:
This whole issue is on hold now due to not having time in any team and if we merge this, we'll break our main product (it needs to get the corresponding PR finished and merged)...

sumeetattree:
Ah, understood. Thanks, I'll try something at my end.

PR #297, opened 2021-06-01
Implement our own LRU for MySQL

pimeys:
We get more control over the implementation and can detect and fix
possible statement leaks more easily.

First part on fixing https://github.com/prisma/prisma/issues/6872


tomhoule (approved):
nice!

PR #299, opened 2021-06-10
feat(mysql): pass prefer_socket to mysql driver

pantharshit00:
This might fix https://github.com/prisma/prisma/issues/6207

Made a quick PR to pass `prefer_socket` to mysql_async so that users that want to disable `SELECT @@socket` call which the driver makes can do so.

### DIFF
--- a/src/connector/mysql.rs
+++ b/src/connector/mysql.rs
@@ -120,6 +120,11 @@ impl MysqlUrl {
         self.query_params.socket_timeout
     }
 
+    /// Prefer socket connection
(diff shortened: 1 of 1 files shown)


pantharshit00:
Hmm, clippy is throwing warnings for the files which I didn't touch. Maybe new version of clippy introduced these lints?

pimeys (approved):
Looks like a right thing to do https://docs.rs/mysql_async/0.26.1/mysql_async/struct.Opts.html#method.prefer_socket

Please inform @janpio about a new parameter. He likes to know.

PR #301, opened 2021-06-17
Optionally expose postgres driver

tomhoule:
This lets us drop down to a lower level API when we need to, at little
maintenance cost. The main use case for this in the Prisma Migration
Engine is to access the full error objects.

I want to extend this to other connectors if this pans out as expected
on postgres.

### DIFF
--- a/src/connector/postgres.rs
+++ b/src/connector/postgres.rs
@@ -26,6 +26,11 @@ use url::Url;
 
 pub(crate) const DEFAULT_SCHEMA: &str = "public";
 
+/// The underlying postgres driver. Only available with the `expose-drivers`
+/// Cargo feature.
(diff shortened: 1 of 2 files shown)


src/connector/postgres.rs
```diff
@@ -510,6 +515,11 @@ impl PostgreSql {
         })
     }
 
+    #[cfg(feature = "expose-drivers")]
```
tomhoule, line 518: Document this? Or wait until we are sure we want this API?

pimeys: No. Reason about the need in rustdoc here to remind us later about the use.

pimeys (approved):
Docs and go

tomhoule:
:+1:

Wrote and pushed some more docstrings.

PR #304, opened 2021-07-02
Catch error on MySQL foreign key creation

pimeys:
Part of: https://github.com/prisma/prisma/issues/7384

### DIFF
--- a/src/connector/mysql/error.rs
+++ b/src/connector/mysql/error.rs
@@ -33,6 +33,17 @@ impl From<my::Error> for Error {
 
                 builder.build()
             }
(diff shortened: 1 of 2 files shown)


src/connector/mysql/error.rs
```diff
@@ -33,6 +33,17 @@ impl From<my::Error> for Error {
 
                 builder.build()
             }
+            my::Error::Server(ServerError { ref message, code, .. }) if code == 1317 => {
```
tomhoule, line 36: Sooo my undenstanding is that the error codes are different between MySQL and Vitess. Are we sure they don't overlap?

pimeys:
closing in favor of solving it in prisma

PR #306, opened 2021-08-06
feat(postgres): support full-text search on Postgres

Weakky:
## Overview

Adds `text_search()` function and `.matches()` comparable to support full-text search on Postgres. Check out rustdocs to see the usages.

### DIFF
--- a/src/ast/compare.rs
+++ b/src/ast/compare.rs
@@ -44,10 +44,10 @@ pub enum Compare<'a> {
     /// Raw comparator, allows to use an operator `left <raw> right` as is,
     /// without visitor transformation in between.
     Raw(Box<Expression<'a>>, Cow<'a, str>, Box<Expression<'a>>),
-    // All json related comparators
+    /// All json related comparators
     #[cfg(all(feature = "json", any(feature = "postgresql", feature = "mysql")))]
     JsonCompare(JsonCompare<'a>),
-    // `left AGAINST (right)
+    /// `left` @@ to_tsquery(`value`)
     #[cfg(feature = "postgresql")]
     Matches(Box<Expression<'a>>, Cow<'a, str>),
 }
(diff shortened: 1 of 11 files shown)


src/ast/compare.rs
```diff
@@ -44,9 +44,12 @@ pub enum Compare<'a> {
     /// Raw comparator, allows to use an operator `left <raw> right` as is,
     /// without visitor transformation in between.
     Raw(Box<Expression<'a>>, Cow<'a, str>, Box<Expression<'a>>),
-    #[cfg(all(feature = "json", any(feature = "postgresql", feature = "mysql")))]
     // All json related comparators
```
pimeys, line 47: Docstring is `///`.

src/ast/compare.rs
```diff
@@ -44,9 +44,12 @@ pub enum Compare<'a> {
...
+    #[cfg(all(feature = "json", any(feature = "postgresql", feature = "mysql")))]
     JsonCompare(JsonCompare<'a>),
+    // `left AGAINST (right)
```
pimeys, line 50: Also, needs to be `///` and needs proper docs.

src/ast/compare.rs
```diff
@@ -770,6 +773,23 @@ pub trait Comparable<'a> {
     where
         T: Into<JsonType>;
 
+    /// Tests if a full-text search matches a certain query. Use it in combination with the `text_search()` function
+    /// ```rust
+    /// # use quaint::{ast::*, visitor::{Visitor, Postgres}};
+    /// # fn main() -> Result<(), quaint::error::Error> {
+    /// let search: Expression = text_search(&vec![Column::from("name"), Column::from("ingredients")]).into();
+    /// let query = Select::from_table("recipes").so_that(search.matches("chicken"));
+    /// let (sql, params) = Postgres::build(query)?;
```
pimeys, line 782: Docs flow better if adding one whitespace after this line.

src/ast/function/search.rs
```diff
@@ -0,0 +1,31 @@
...
+/// ```rust
+/// # use quaint::{ast::*, visitor::{Visitor, Postgres}};
+/// # fn main() -> Result<(), quaint::error::Error> {
+/// let search: Expression = text_search(&vec![Column::from("name"), Column::from("ingredients")]).into();
+/// let query = Select::from_table("recipes").so_that(search.matches("chicken"));
+/// let (sql, params) = Postgres::build(query)?;
+/// assert_eq!("SELECT \"recipes\".* FROM \"recipes\" WHERE to_tsvector(\"name\"|| ' ' ||\"ingredients\") @@ to_tsquery($1)", sql);
+/// assert_eq!(params, vec![Value::from("chicken")]);
+/// # Ok(())    
+/// # }
```
pimeys, line 20: I'm actually thinking how the functions here always return a `Function`, but should we actually just always return an `Expression`. Do we ever need the `Function` type returned anywhere?

This would simplify the code a lot:

```rust
let search = text_search(&[Column::from("name"), Column::from("ingredients")]);
let query = Select::from_table("recipes").so_that(search.matches("chicken"));
let (sql, _) = Postgres::build(query)?;

assert_eq!(
    "SELECT \"recipes\".* FROM \"recipes\" WHERE to_tsvector(\"name\"|| ' ' ||\"ingredients\") @@ to_tsquery($1)",
    sql
);
```

See also how the example can be written in a simpler way. No need for `vec!` if you just take a slice, and the pacing and line length should be considered more carefully.

pimeys (commented):
Only PostgreSQL this time? Some nitpicks and maybe a refactoring you could do after this is merged by seeing if the functions could already return an `Expression` instead of a `Function`.

PR #310, opened 2021-08-13
fix(mysql): numeric comparison with json_extract

Weakky:
## Overview

Fixes JSON numeric comparison for MySQL


pimeys (approved):
This is how we solve database-related conversions anyhow. It's the right way to do it.

PR #313, opened 2021-08-19
feat(pg): add not_matches for full-text search

Weakky:
## Overview

Supports `.not_matches()`, the inverse of `.matches()` for full-text search on Postgres


pimeys (approved):
no complaints this time.

PR #314, opened 2021-08-19
chore: sort query params

janpio:


src/pooled.rs
```diff
@@ -76,14 +82,17 @@
 //! - `sslaccept` either `strict` or `accept_invalid_certs`. If strict, the
 //!   certificate needs to be valid and in the CA certificates.
 //!   `accept_invalid_certs` accepts any certificate from the server and can
-//!   lead to weakened security. Defaults to `strict`.
+//!   lead to weakened security. Defaults to `accept_invalid_certs`.
```
janpio, line 85: I think this was documented wrong until now.

src/single.rs
```diff
@@ -95,34 +99,38 @@ impl Quaint {
...
+    ///   disable TLS completely, including sending login credentials as plaintext.
     /// - `trustServerCertificate` if set to `true`, accepts any kind of certificate
     ///   from the server.
     /// - `socketTimeout` defined in seconds. If set, a query will return a
     ///   `Timeout` error if it fails to resolve before given time.
     /// - `connectTimeout` defined in seconds (default: 5). Connecting to a
     ///   database will return a `ConnectTimeout` error if taking more than the
     ///   defined value.
-    /// - `connectionLimit` defines the maximum number of connections opened to the
-    ///   database.
```
janpio, line 123: No `connectionLimit` in single I assume.

PR #315, opened 2021-08-20
fix(mariadb): json gt/gte/lt/lte comparisons

Weakky:
## Overview

Follow-up PR to https://github.com/prisma/quaint/pull/311

It turns out `CAST(? as JSON)` is not supported by MariaDB. We work around this limitation by no longer rendering a JSON value compared to `JSON_EXTRACT()` as JSON, but as its parsed type.

As a result though, it is no longer possible to perform numeric comparisons with `JSON_EXTRACT()` as the left/right and a JSON value as the left/right member **that's NOT a number or a string**.

## Examples

```diff
SELECT ... FROM ... WHERE
- JSON_EXTRACT(?, ?) > CAST("1" AS JSON)
+ JSON_EXTRACT(?, ?) > 1
```

```diff
SELECT ... FROM ... WHERE
- JSON_EXTRACT(?, ?) > CAST('"hello"' AS JSON)
+ JSON_EXTRACT(?, ?) > "hello"
```

## Additional notes

Since we added MySQL8 and MariaDB to the list of tested databases, some failing tests have appeared. I created an issue to track this: https://github.com/prisma/quaint/issues/318

### DIFF
--- a/src/tests/query.rs
+++ b/src/tests/query.rs
@@ -1204,7 +1204,9 @@ async fn deletes(api: &mut dyn TestApi) -> crate::Result<()> {
     Ok(())
 }
 
-#[test_each_connector(tags("mysql"))]
+// Figure out why it doesn't work on MariaDB
+// Error { kind: QueryError(Server(ServerError { code: 1115, message: "Unknown character set: 'gb18030'", state: "42000" })), original_code: Some("1115"), original_message: Some("Unknown character set: 'gb18030'") }
+#[test_each_connector(tags("mysql"), ignore("mysql_mariadb"))]
 async fn text_columns_with_non_utf8_encodings_can_be_queried(api: &mut dyn TestApi) -> crate::Result<()> {
     let table = api
         .create_table("id integer auto_increment primary key, value varchar(100) character set gb18030")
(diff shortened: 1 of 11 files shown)


src/tests/query.rs
```diff
@@ -1955,6 +1955,20 @@ async fn coalesce_fun(api: &mut dyn TestApi) -> crate::Result<()> {
...
+        Value::Text(Some(text)) => {
+            let json: serde_json::Value =
+                serde_json::from_str(&text).expect(format!("expected parsable text to json, found {}", text).as_str());
+
+            Some(json)
+        }
+        Value::Json(Some(json)) => Some(json),
+        _ => None,
+    }
+}
```
Weakky, line 1970: I ended up using an external helper and not adding a match to `Value::as_json()` because it's a const method and `serde_json::from_str` is not so it cannot work. And changing the const fn to a simple fn I believe is a breaking change that wasn't worth it.

src/visitor/mysql.rs
```diff
@@ -32,16 +32,47 @@ impl<'a> Mysql<'a> {
...
+                    let msg = format!("Expected JSON string or number, found: {}", x);
+                    let kind = ErrorKind::conversion(msg.clone());
+
+                    let mut builder = Error::builder(kind);
+                    builder.set_original_message(msg);
+
+                    Err(builder.build())
+                }
+            }
+        }
```
Weakky, line 58: This is the actual fix. We no longer use `CAST (? AS JSON)` but instead render the values deserialized, which works both on MySQL & MariaDB

src/visitor/mysql.rs
```diff
@@ -32,16 +32,47 @@ impl<'a> Mysql<'a> {
...
+
         match (left, right) {
+            #[cfg(feature = "json")]
             (left, right) if left.is_json_value() && right.is_json_extract_fun() => {
-                self.surround_with("CAST(", " AS JSON)", |s| s.visit_expression(left))?;
+                let quaint_value = json_to_quaint_value(left.into_json_value().unwrap())?;
```
Weakky, line 63: Usage of the function above

PR #319, opened 2021-08-24
fix(mariadb): json array filters

Weakky:


src/ast/function/json_extract_array.rs
```diff
@@ -0,0 +1,39 @@
...
+pub(crate) fn json_extract_first_array_elem<'a, E>(expr: E) -> Function<'a>
+where
+    E: Into<Expression<'a>>,
+{
+    let fun = JsonExtractFirstArrayElem {
+        expr: Box::new(expr.into()),
+    };
+
+    fun.into()
+}
```
Weakky, line 39: I created two new _private_ functions to enable using the existing JSON equality code paths. See the comment below to see how they are used 👇

src/ast/expression.rs
```diff
@@ -466,31 +466,39 @@ impl<'a> Comparable<'a> for Expression<'a> {
...
 
     #[cfg(all(feature = "json", any(feature = "postgresql", feature = "mysql")))]
     fn json_array_not_ends_into<T>(self, item: T) -> Compare<'a>
     where
         T: Into<Expression<'a>>,
     {
-        Compare::JsonCompare(JsonCompare::ArrayNotEndsInto(Box::new(self), Box::new(item.into())))
+        let array_ends_into: Expression = json_extract_last_array_elem(self).into();
+
+        Compare::NotEquals(Box::new(array_ends_into), Box::new(item.into()))
```
Weakky, line 501: The private functions are used here 👆.

Instead of having special `JsonCompare`s, we rely on the existing `Compare::Equals` & `Compare::NotEquals`.

The private functions help us _wrap_ the left side of the equal into whatever's needed to get the first/last item of an array.

src/ast/compare.rs
```diff
@@ -59,10 +59,6 @@ pub enum Compare<'a> {
 pub enum JsonCompare<'a> {
     ArrayContains(Box<Expression<'a>>, Box<Expression<'a>>),
     ArrayNotContains(Box<Expression<'a>>, Box<Expression<'a>>),
-    ArrayBeginsWith(Box<Expression<'a>>, Box<Expression<'a>>),
-    ArrayNotBeginsWith(Box<Expression<'a>>, Box<Expression<'a>>),
-    ArrayEndsInto(Box<Expression<'a>>, Box<Expression<'a>>),
-    ArrayNotEndsInto(Box<Expression<'a>>, Box<Expression<'a>>),
```
Weakky, line 65: As explained above, this means we no longer need these `JsonCompare` variants as we now rely on `Compare::Equals` & `Compare::NotEquals`.

src/visitor.rs
```diff
@@ -872,10 +872,6 @@ pub trait Visitor<'a> {
             Compare::JsonCompare(json_compare) => match json_compare {
                 JsonCompare::ArrayContains(left, right) => self.visit_json_array_contains(*left, *right, false),
                 JsonCompare::ArrayNotContains(left, right) => self.visit_json_array_contains(*left, *right, true),
-                JsonCompare::ArrayBeginsWith(left, right) => self.visit_json_array_begins_with(*left, *right, false),
-                JsonCompare::ArrayNotBeginsWith(left, right) => self.visit_json_array_begins_with(*left, *right, true),
-                JsonCompare::ArrayEndsInto(left, right) => self.visit_json_array_ends_into(*left, *right, false),
-                JsonCompare::ArrayNotEndsInto(left, right) => self.visit_json_array_ends_into(*left, *right, true),
```
Weakky, line 878: As explained above though, we could as well keep the `JsonCompare` variants and create the `Compare::Equals|NotEquals` variants in here if that's preferable for visibility.

src/tests/test_api.rs
```diff
@@ -25,6 +27,7 @@ pub trait TestApi {
     async fn create_index(&mut self, table: &str, columns: &str) -> crate::Result<String>;
 
     fn system(&self) -> &'static str;
+    fn connector_tag(&self) -> Tags;
```
Weakky, line 30: Last thing: unfortunately, `JSON_CONTAINS` doesn't behave the same on MariaDB as it does on MySQL. This was causing issues on test assertions. I, therefore, added a new method to enable getting the current tag of the connector. See the comment below how it's being used 👇🏻

src/tests/query.rs
```diff
@@ -2111,36 +2113,52 @@ async fn json_array_contains_fun(api: &mut dyn TestApi) -> crate::Result<()> {
...
+            value_into_json(&res.next().unwrap()["obj"])
+        );
+        assert_eq!(None, res.next());
+    } else {
+        assert_eq!(
+            Some(serde_json::json!({ "a": { "b": [[1, 2], [3, 4]] } })),
+            value_into_json(&res.next().unwrap()["obj"])
+        );
+        assert_eq!(None, res.next());
+    }
```
Weakky, line 2161: This is where we branch the assertions in the case of MariaDB

pimeys (approved):
lgtm

Hopefully no more weird JSON edge cases after this :)

PR #333, opened 2021-11-12
Add the pool timeout duration to the pool timeout error

garrensmith:
This is part of the fix for https://github.com/prisma/prisma/issues/9662

### DIFF
--- a/src/pooled.rs
+++ b/src/pooled.rs
@@ -465,7 +465,11 @@ impl Quaint {
             Ok(conn) => conn,
             Err(mobc::Error::Timeout) => {
                 let state = self.inner.state().await;
-                return Err(Error::builder(ErrorKind::pool_timeout(state.max_open, state.in_use)).build());
+                // We can use unwrap here because a pool timeout has to be set to use a connection pool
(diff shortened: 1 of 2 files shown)


src/pooled.rs
```diff
@@ -465,7 +465,10 @@ impl Quaint {
             Ok(conn) => conn,
             Err(mobc::Error::Timeout) => {
                 let state = self.inner.state().await;
-                return Err(Error::builder(ErrorKind::pool_timeout(state.max_open, state.in_use)).build());
+                let timeout_duration = self.pool_timeout.unwrap(); // We can use unwrap here because a pool timeout had to be set to get here
```
dpetrick, line 468: It took me a second to understand the unwrap reason. You can't run into a timeout error without setting a timeout, so the timeout option must be set. Potentially rephrase it, but not a blocker.

dpetrick (approved):

src/pooled.rs
garrensmith: Cool. I've updated it a bit to make it clearer.

PR #336, opened 2021-12-03
Prepare renaming of main branch from master to main

tomhoule:
The GitHub workflow is changed to use a fixed version of the toolchain through the standard [rust-toolchain](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) format. That makes CI builds more reproducible.  

Also renames master to main.


rust-toolchain
```diff
@@ -0,0 +1 @@
+1.56.0
```
pimeys, line 1: ?

pimeys (approved):

rust-toolchain
janpio: https://github.com/prisma/tiberius/pull/185/files#r764314572

tomhoule:
See my comment on the tiberius PR, let's experiment with that, but change things if it's additional maintenance burden (totally legit complaint)

PR #337, opened 2021-12-21
MySql Full text support

garrensmith:
Add the ability to query a MySql full text index.

There is also a fix to how we perform text indexes with Postgres.

### DIFF
--- a/src/visitor/mysql.rs
+++ b/src/visitor/mysql.rs
@@ -441,19 +441,38 @@ impl<'a> Visitor<'a> for Mysql<'a> {
         Ok(())
     }
 
-    #[cfg(feature = "postgresql")]
-    fn visit_text_search(&mut self, _text_search: crate::prelude::TextSearch<'a>) -> visitor::Result {
-        unimplemented!("Full-text search is not yet supported on MySQL")
+    #[cfg(any(feature = "postgresql", feature = "mysql"))]
+    fn visit_text_search(&mut self, text_search: crate::prelude::TextSearch<'a>) -> visitor::Result {
+        let len = text_search.exprs.len();
+        self.surround_with("MATCH (", ")", |s| {
+            for (i, expr) in text_search.exprs.into_iter().enumerate() {
+                s.visit_expression(expr)?;
+
+                if i < (len - 1) {
+                    s.write(",")?;
+                }
+            }
+
+            Ok(())
+        })
     }
 
-    #[cfg(feature = "postgresql")]
-    fn visit_matches(
-        &mut self,
-        _left: Expression<'a>,
-        _right: std::borrow::Cow<'a, str>,
-        _not: bool,
-    ) -> visitor::Result {
-        unimplemented!("Full-text search is not yet supported on MySQL")
+    #[cfg(any(feature = "postgresql", feature = "mysql"))]
+    fn visit_matches(&mut self, left: Expression<'a>, right: std::borrow::Cow<'a, str>, not: bool) -> visitor::Result {
+        if not {
+            self.write("(NOT ")?;
+        }
+
+        self.visit_expression(left)?;
+        self.surround_with("AGAINST (", " IN BOOLEAN MODE)", |s| {
+            s.visit_parameterized(Value::text(right))
+        })?;
+
+        if not {
+            self.write(")")?;
+        }
+
+        Ok(())
     }
 
     #[cfg(all(feature = "json", any(feature = "postgresql", feature = "mysql")))]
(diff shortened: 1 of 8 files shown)


src/visitor/mysql.rs
```diff
@@ -441,19 +441,38 @@ impl<'a> Visitor<'a> for Mysql<'a> {
...
-
-    #[cfg(feature = "postgresql")]
-    fn visit_matches(
-        &mut self,
-        _left: Expression<'a>,
-        _right: std::borrow::Cow<'a, str>,
-        _not: bool,
-    ) -> visitor::Result {
-        unimplemented!("Full-text search is not yet supported on MySQL")
+    #[cfg(any(feature = "postgresql", feature = "mysql"))]
```
tomhoule, line 444: It's weird that we need this for the crate to compile, but I think I understand why.

tomhoule (approved):
I haven't used these kinds of functions so I can't review the exact shape of the generated SQL, but it does look good :+1: Nice fix on postgres.

src/visitor/mysql.rs
garrensmith: I'm open to better ways of doing that

Weakky: IIUC, we don't need those for the QE. We need those because Quaint is open-source and we want to offer fine-grained APIs based on which features you've enabled. QE side we always enable all of them anyway since we don't compile different QEs for different connectors.
This might turn out to be useful in the future though.

tomhoule: yep, that's the reasoning. I wonder if there would be ways to have these feature gates in fewer places.

tomhoule: maybe group methods in modules and have feature gates on those

tomhoule: Definitely a larger project that doesn't fit in this PR though

src/visitor/postgres.rs
```diff
@@ -365,12 +365,12 @@ impl<'a> Visitor<'a> for Postgres<'a> {
...
     fn visit_text_search(&mut self, text_search: crate::prelude::TextSearch<'a>) -> visitor::Result {
         let len = text_search.exprs.len();
-        self.surround_with("to_tsvector(", ")", |s| {
+        self.surround_with("to_tsvector(concat_ws(' ', ", "))", |s| {
             for (i, expr) in text_search.exprs.into_iter().enumerate() {
                 s.visit_expression(expr)?;
 
                 if i < (len - 1) {
-                    s.write("|| ' ' ||")?;
+                    s.write(",")?;
```
janpio, line 373: This unrelated change in here looks scary.

garrensmith: Its not unrelated. This is related to switching to using `concat_ws`

tomhoule: How I understand this: the generated query fragment went from `to_tsvector('a' || ' ' || 'b' || ' ' || 'c')` (`||` is the string concatenation operator on postgres), to `to_tsvector(concat_ws(' ', 'a', 'b', 'c'))`, which is better since if any of a, b, or c is null, you will still get the other ones in the concatenated string, but not so with `||`.

tomhoule: (so if `a` is NULL, `a || ' ' || 'b'` would be NULL, but `concat_ws(' ', a, 'b')` would be ` b`)

garrensmith: Good explanation Tom, thanks. That is correct.

garrensmith: For example, based on our tests:
Original columns:
```
postgres=# SELECT "id", "fieldA", "fieldC" from "public"."TestModel";
 id |     fieldA     | fieldC  
----+----------------+---------
  1 | Chicken Masala | 
  2 | Chicken Curry  | 
  3 | Caesar Salad   | Chicken
  4 | Beef Sandwich  | Beef
  ```
  
```
SELECT "id", "fieldA", "fieldC",  to_tsvector("fieldA" || ' ' ||  "fieldC") from "public"."TestModel";
 id |     fieldA     | fieldC  |           to_tsvector            
----+----------------+---------+----------------------------------
  1 | Chicken Masala |         | 
  2 | Chicken Curry  |         | 
  3 | Caesar Salad   | Chicken | 'caesar':1 'chicken':3 'salad':2
  4 | Beef Sandwich  | Beef    | 'beef':1,3 'sandwich':2
  ```
  
  Then using `concat_ws`
  ```
  postgres=# SELECT "id", "fieldA", "fieldC",  to_tsvector(concat("fieldA","fieldC")) from "public"."TestModel";
 id |     fieldA     | fieldC  |         to_tsvector         
----+----------------+---------+-----------------------------
  1 | Chicken Masala |         | 'chicken':1 'masala':2
  2 | Chicken Curry  |         | 'chicken':1 'curri':2
  3 | Caesar Salad   | Chicken | 'caesar':1 'saladchicken':2
  4 | Beef Sandwich  | Beef    | 'beef':1 'sandwichbeef':2
  ```

garrensmith: Notice how when using `||` the first two rows are missing all text

janpio: Looked unrelated as PR is about MySQL, change is in PostgreSQL file.

Issue #338, opened 2021-12-21
PostgreSQL: Support `options` connection string parameter


janpio:
PostgreSQL supports an `options` parameter for connection strings:

> Specifies command-line options to send to the server at connection start. For example, setting this to -c geqo=off sets the session's value of the geqo parameter to off. Spaces within this string are considered to separate command-line arguments, unless escaped with a backslash (\); write \\ to represent a literal backslash.

https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-OPTIONS

Cockroach Cloud for example uses that to submit the cluster name when a connection is created: `...?options=--cluster%3D{{CLUSTER_NAME}}`

Our PostgreSQL driver Tokio-Postgres also support this parameter:

> `options` - Command line options used to configure the server.

https://docs.rs/tokio-postgres/latest/tokio_postgres/config/struct.Config.html

Right now it seems we are filtering this value out of the connection string, or at least not forwarding it to Tokio-Postgres: https://prisma.github.io/quaint/src/quaint/connector/postgres.rs.html#267

It would be great if we could enable Prisma to use the `options` parameter in PostgreSQL connection strings.

PR #339, opened 2021-12-22
feat(Postgres): Support options field in connection string

garrensmith:
Add support for the `options` field in the postgres connection string https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-OPTIONS

This resolves https://github.com/prisma/quaint/issues/338

### DIFF
--- a/src/connector/postgres.rs
+++ b/src/connector/postgres.rs
@@ -264,6 +264,10 @@ impl PostgresUrl {
         }
     }
 
+    pub(crate) fn options(&self) -> &str {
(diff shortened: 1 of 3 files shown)


tomhoule (approved):
I didn't know about that param. Command line flags in URL params is "interesting", I've never seen that :p But the implementation is good and it matches what the postgres docs say, as far as I can tell. Nice!

tomhoule (approved):
(approving again for the docs)

PR #340, opened 2021-12-23
fix(postgres) Only pass options params when set

garrensmith:
Only pass the options field through to postgres when it
has been set.

This fixes https://github.com/prisma/prisma/issues/10841


tomhoule (approved):
better!

tomhoule:
approved

PR #341, opened 2021-12-23
feat(mysql): Add text search relevance to mysql

garrensmith:
Adds `visit_text_search_relevance` support to mysql and refactors the
original postgres only support to allow any connector to be able to
implement it.

### DIFF
--- a/src/visitor/mssql.rs
+++ b/src/visitor/mssql.rs
@@ -620,6 +620,14 @@ impl<'a> Visitor<'a> for Mssql<'a> {
         unimplemented!("Full-text search is not yet supported on MSSQL")
     }
 
+    #[cfg(feature = "postgresql")]
+    fn visit_text_search_relevance(
+        &mut self,
+        _text_search_relevance: crate::prelude::TextSearchRelevance<'a>,
+    ) -> visitor::Result {
+        unimplemented!("Full-text search is not yet supported on MSSQL")
+    }
+
     #[cfg(all(feature = "json", any(feature = "postgresql", feature = "mysql")))]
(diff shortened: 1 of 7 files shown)


src/visitor/mssql.rs
```diff
@@ -620,6 +620,14 @@ impl<'a> Visitor<'a> for Mssql<'a> {
         unimplemented!("Full-text search is not yet supported on MSSQL")
     }
 
+    #[cfg(feature = "postgresql")]
```
dpetrick, line 623: This looks off, why is there a postgres feature flag in mssql?

src/visitor/mysql.rs
```diff
@@ -475,6 +475,18 @@ impl<'a> Visitor<'a> for Mysql<'a> {
         Ok(())
     }
 
+    #[cfg(any(feature = "postgresql", feature = "mysql"))]
```
dpetrick, line 478: Do you really need these feature flags? The whole mysql struct is gated by the `mysql` feature flag already.

src/visitor/postgres.rs
```diff
@@ -394,6 +394,31 @@ impl<'a> Visitor<'a> for Postgres<'a> {
 
         Ok(())
     }
+
+    #[cfg(any(feature = "postgresql", feature = "mysql"))]
```
dpetrick, line 398: Same as the above comment.

src/visitor/mssql.rs
garrensmith: This is needed for when we have `features=all`

dpetrick (approved):

PR #343, opened 2022-01-16
don't panic if decimals don't fit

snf:
Hi, I'm hitting this in my prisma code if the big decimal is too large.
Found it with the number 6.673310573781466e+51 . Adding a check in my code but, regardless, I don't think the lib should panic.


pimeys:
No it should not panic. Thanks for the fix. Will merge tomorrow.

snf:
Thank you for the prompt response!

tomhoule (approved):

PR #344, opened 2022-01-17
postgres connector don't panic parsing mantissa

snf:


garrensmith (commented):
This looks good. Any chance you could add a test for this?

snf:
I had a look into making a test but couldn't see an easy way to add one as the `test_type!` ones are for success (and this is a fail).
Anyway, I don't think there is a risk here because this only changes a panic into an error.

Weakky:
Hey @snf,

I created a [follow-up PR here](https://github.com/prisma/quaint/pull/379) which includes your commit.

I'll close this one in favor of this other one now. Thanks for contributing! 🙏

PR #346, opened 2022-01-26
Support Uuid_to_bin in MySQL

cprieto:
MySQL is the only main used database returning Uuid as a string, so they created an additional function, uuid_to_bin, to convert the string value to binary. This is a "feature" required for https://github.com/prisma/prisma/issues/7010.

This is the clearer of all the options... Another set of ideas could be:

 - Adding a `Literal(str)` function passing _anything_ you want to use as a function
 - Creating a new value type, `Unquoted` which is not going to be quoted as other `RawValue`s
 - Adding a new `Expression`, literal, similar to first option but using expression directly

At the end this was the clearer, of course it has a few tradeoffs:

 - It does not work in MySQL 5.x (uuid_to_bin)
 
This applies as well for the MySQL function `uuid` which generates a string instead of a binary

### DIFF
--- a/src/ast/function.rs
+++ b/src/ast/function.rs
@@ -17,6 +17,9 @@ mod search;
 mod sum;
 mod upper;
 
+#[cfg(feature = "mysql")]
+mod uuid;
+
 pub use aggregate_to_string::*;
 pub use average::*;
 pub use coalesce::*;
@@ -35,6 +38,9 @@ pub use row_to_json::*;
 pub use search::*;
 pub use sum::*;
 pub use upper::*;
+
+#[cfg(feature = "mysql")]
+pub use self::uuid::*;
 
 use super::{Aliasable, Expression};
 use std::borrow::Cow;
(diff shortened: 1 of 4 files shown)


src/ast/function.rs
```diff
@@ -71,6 +71,30 @@ pub(crate) enum FunctionType<'a> {
...
+
+/// let query = Select::default().value(uuid_to_bin());
+/// let (sql, _) = Mysql::build(query)?;
+///
+/// assert_eq!("SELECT uuid_to_bin(uuid())", sql);
+/// # Ok(())
+/// # }
+/// ```
+#[cfg(any(feature = "mysql"))]
+pub fn uuid_to_bin() -> Expression<'static> {
```
garrensmith, line 91: I don't know quaint super well. But why does this have to be an expression over something like:
```
function!(UUIDToBin);
```

garrensmith: Ah ok. That makes sense.

cprieto: That is only for expression structs, not enum members, I got super confussed by that as well

src/ast/function.rs
```diff
@@ -71,6 +71,30 @@ pub(crate) enum FunctionType<'a> {
     TextSearch(TextSearch<'a>),
     #[cfg(any(feature = "postgresql", feature = "mysql"))]
     TextSearchRelevance(TextSearchRelevance<'a>),
+    #[cfg(any(feature = "mysql"))]
```
Weakky, line 74: This could be simplified to: `cfg(feature = "mysql")`

src/tests/query.rs
```diff
@@ -2871,3 +2871,29 @@ async fn delete_comment(api: &mut dyn TestApi) -> crate::Result<()> {
...
 }
+
+#[cfg(feature = "mysql")]
+#[test_each_connector(tags("mysql8"))]
+async fn generate_binary_uuid(api: &mut dyn TestApi) -> crate::Result<()> {
+    let select = Select::default().value(uuid_to_bin());
+    let res = api.conn().select(select).await?.into_single()?;
+    let val = res.into_single()?;
+
+    // Just check is a byte and it has some value, it is a uuid!
```
dpetrick, line 2882:
```suggestion
    // If it is a byte type and has a value, it's a generated UUID.
```

src/tests/query.rs
```diff
@@ -2871,3 +2871,29 @@ async fn delete_comment(api: &mut dyn TestApi) -> crate::Result<()> {
...
+}
+
+#[cfg(feature = "mysql")]
+#[test_each_connector(tags("mysql"))]
+async fn generate_native_uuid(api: &mut dyn TestApi) -> crate::Result<()> {
+    let select = Select::default().value(native_uuid());
+    let res = api.conn().select(select).await?.into_single()?;
+    let val = res.into_single()?;
+
+    // Just check is a text and it has some value, it is a string uuid!
```
dpetrick, line 2895:
```suggestion
    // If it is a text type and has a value, it's a generated string UUID.
```

dpetrick (approved):

PR #358, opened 2022-04-20
feat: pass type-hint to PG prepared statement

Weakky:
## Overview


src/connector/postgres.rs
```diff
@@ -619,18 +620,50 @@ impl TransactionCapable for PostgreSql {}
...
+
+            Ok(result)
+        })
+        .await
+    }
+
+    #[tracing::instrument(skip(self, params))]
+    async fn query_raw_typed(&self, sql: &str, params: &[Value<'_>]) -> crate::Result<ResultSet> {
+        metrics::query("postgres.query_raw", sql, params, move || async move {
+            let param_types = conversion::params_to_types(params);
```
pimeys, line 665: could we skip this if we already have the statement in the cache? it's a lot of cycles in the worst case...

src/connector/postgres/conversion.rs
```diff
@@ -29,6 +29,92 @@ pub fn conv_params<'a>(params: &'a [Value<'a>]) -> Vec<&'a (dyn types::ToSql + S
...
+                Value::Text(_) => PostgresType::TEXT,
+                // Enums are special types, we can't statically infer them, so we let PG infer it
+                Value::Enum(_) => PostgresType::UNKNOWN,
+                Value::Bytes(_) => PostgresType::BYTEA,
+                Value::Boolean(_) => PostgresType::BOOL,
+                Value::Char(_) => PostgresType::CHAR,
+                #[cfg(feature = "bigdecimal")]
+                Value::Numeric(_) => PostgresType::NUMERIC,
+                #[cfg(feature = "json")]
+                Value::Json(_) => PostgresType::JSON,
```
pimeys, line 57: I'm worried about this here, you should always prefer jsonb if you can, json is only for function return values. Did you have some tests failing if you'd use jsonb here?

src/connector/postgres/conversion.rs
```diff
@@ -29,6 +29,92 @@ pub fn conv_params<'a>(params: &'a [Value<'a>]) -> Vec<&'a (dyn types::ToSql + S
...
+                        Value::Text(_) => PostgresType::TEXT_ARRAY,
+                        // Enums are special types, we can't statically infer them, so we let PG infer it
+                        Value::Enum(_) => PostgresType::UNKNOWN,
+                        Value::Bytes(_) => PostgresType::BYTEA_ARRAY,
+                        Value::Boolean(_) => PostgresType::BOOL_ARRAY,
+                        Value::Char(_) => PostgresType::CHAR_ARRAY,
+                        #[cfg(feature = "bigdecimal")]
+                        Value::Numeric(_) => PostgresType::NUMERIC_ARRAY,
+                        #[cfg(feature = "json")]
+                        Value::Json(_) => PostgresType::JSON_ARRAY,
```
pimeys, line 99: Maybe jsonb_array?

src/tests/query.rs
```diff
@@ -3081,3 +3081,98 @@ async fn generate_native_uuid(api: &mut dyn TestApi) -> crate::Result<()> {
 
     Ok(())
 }
+
+#[test_each_connector(tags("postgresql"))]
+async fn query_raw_typed_numeric(api: &mut dyn TestApi) -> crate::Result<()> {
```
pimeys, line 3086: Test with json!

pimeys (approved):
accepting but please review some of the comments before merging...

src/connector/postgres.rs
Weakky: Yeah, that was stupid from me 🤦‍♂️

PR #361, opened 2022-05-04
fix: allow nulls in postgres scalar lists

Weakky:
## Overview

Fixes https://github.com/prisma/prisma/issues/12367


src/connector/postgres/conversion.rs
```diff
@@ -646,10 +670,11 @@ impl<'a> ToSql for Value<'a> {
                 }),
                 #[cfg(feature = "uuid")]
                 (Value::Array(values), &PostgresType::UUID_ARRAY) => values.as_ref().map(|values| {
-                    let parsed_uuid: Vec<Uuid> = values
+                    let parsed_uuid: Vec<Option<Uuid>> = values
                         .iter()
-                        .filter_map(|v| v.to_string().and_then(|v| v.parse().ok()))
```
Weakky, line 651: `Value::to_string()` only returns `Some(x)` for the `Value::String` and `Value::Bytes` variants. Combined with `filter_map`, this had two effects:
- `Value::Uuid` were not inserted
- `String` and `Bytes` that could not be coerced to `Uuid` were silently ignored

Same goes for `INET_ARAY` and  `BIT_ARRAY`

src/connector/postgres/conversion.rs
```diff
@@ -646,10 +670,11 @@ impl<'a> ToSql for Value<'a> {
...
-                        .collect();
+                        .map(<Option<Uuid>>::try_from)
+                        .collect::<crate::Result<Vec<_>>>()?;
```
Weakky, line 676: `Value::to_string()` only returns `Some(x)` for the `Value::String` and `Value::Bytes` variants. Combined with `filter_map`, this had two effects:
- `Value::Uuid` were not inserted
- `String` and `Bytes` that could not be coerced to `Uuid` were silently ignored

Same goes for `INET_ARAY` and  `BIT_ARRAY`

pimeys (approved):
great.

PR #363, opened 2022-05-08
add two more metrics to quaint

garrensmith:


src/connector/metrics.rs
```diff
@@ -29,6 +29,8 @@ where
     );
 
     histogram!(format!("{}.query.time", tag), start.elapsed());
+    histogram!("query.total_elapsed_time", start.elapsed());
```
tomhoule, line 32: Question because I really don't know: is it the same metric as the previous line, but with a different name?

tomhoule (approved):
Looks good, with a caveat that I don't have much context on metrics.

src/connector/metrics.rs
garrensmith: Yes. It is a more general metric for all queries

PR #365, opened 2022-05-10
Add support for "swapped" UUIDs in MySQL

cprieto:
- MySQL 8 has a "fantastic" feature for index-optimized UUIDs, they are called swapped UUIDs
 - Sadly, they are enabled with a parameter in UUID
 - See https://dev.mysql.com/doc/refman/8.0/en/miscellaneous-functions.html#function_uuid-to-bin
 - "Normal" UUID to bin still as is (equivalent parameter, 0)
 - This generates a swapped UUID for generated ids
 - Will solve (partially) https://github.com/prisma/prisma/issues/12850

### DIFF
--- a/src/ast/function.rs
+++ b/src/ast/function.rs
@@ -80,6 +80,8 @@ pub(crate) enum FunctionType<'a> {
     #[cfg(feature = "mysql")]
     UuidToBin,
     #[cfg(feature = "mysql")]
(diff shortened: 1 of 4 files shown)


janpio (commented):
Looks good, but as I don't really know Rust, just a 👍 as a comment.

dpetrick (approved):

PR #368, opened 2022-05-25
feat: remove query_raw_typed and make it the default

Weakky:
## Overview

- BREAKING: Remove query_raw_typed and make it the default behaviour via query_raw
- Prisma 4 change


Weakky:
`query_raw_typed` will stay for now.

PR #369, opened 2022-05-31
Distinction between int32 and int64 in SQLite

cprieto:
- This is part of https://github.com/prisma/prisma/issues/12784
- Now, `INT` is `int32` and `BIGINT` is `int64` in SQLite
- When _we don't know what type is it_ we are now `int32(None)` (this happens with `NULL` types), used to be `int64(None)`
- Certain operations, for example, `RETURNING` cannot infer the return type, in cases where we cannot infer the return type for an integer it will be set as `int64` (another example is with implicit values)

### DIFF
--- a/src/connector/sqlite/conversion.rs
+++ b/src/connector/sqlite/conversion.rs
@@ -133,7 +133,9 @@ impl<'a> GetRow for SqliteRow<'a> {
         for (i, column) in self.columns().iter().enumerate() {
             let pv = match self.get_ref_unwrap(i) {
                 ValueRef::Null => match column {
-                    c if c.is_int32() | c.is_int64() | c.is_null() => Value::Int64(None),
+                    // NOTE: A value without decl_type would be Int32(None)
+                    c if c.is_int32() | c.is_null() => Value::Int32(None),
+                    c if c.is_int64() => Value::Int64(None),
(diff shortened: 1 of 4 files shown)


src/connector/sqlite/conversion.rs
```diff
@@ -103,6 +105,8 @@ impl TypeIdentifier for Column<'_> {
         }
     }
 
+    // WITH (id) AS pepe (SELECT 1) SELECT id (None) FROM pepe
+
```
Weakky, line 109: You forgot to remove that 😛

src/connector/sqlite/conversion.rs
```diff
@@ -169,6 +176,17 @@ impl<'a> GetRow for SqliteRow<'a> {
                         let dt = chrono::Utc.timestamp_millis(i);
                         Value::datetime(dt)
                     }
+                    c if c.is_int32() => {
+                        if let Ok(converted) = i32::try_from(i) {
+                            Value::int32(converted)
+                        } else {
+                            let msg = format!("Value {} does not fit in an INT column, maybe it is a BIGINT", i);
```
Weakky, line 183: We should give suggestions as to how to fix the error. Something like:

```
Value {} does not fit in an `Int` column. Please try migrating the '{}' column type to `BigInt`.
```

PR #370, opened 2022-06-01
Remove support for sqlite url prefix

pimeys:
Merge only when Prisma 4 is due

Part of: https://github.com/prisma/prisma/issues/11468


tomhoule (approved):

pimeys:
> There is another case I think:
> 
> https://github.com/prisma/quaint/blob/8be4fab7f2f6213487b1d40f2d23a6de982a0cf6/src/single.rs#L134

I didn't save that file before committing...

janpio (changes requested):
see comments

janpio (approved):

PR #379, opened 2022-06-15
fix(pg): support decimal mantissa greater than u128::MAX

Weakky:
## Overview

Follow-up PR to https://github.com/prisma/quaint/pull/344

fixes https://github.com/prisma/prisma/issues/13219


pimeys (approved):
Mantissa madness.

PR #380, opened 2022-06-20
fix(sqlite): float columns should return floats

Weakky:
## Overview

fixes https://github.com/prisma/prisma-engines/issues/2991

Float columns on SQLite are migrated as `REAL`. However, `REAL` columns were coerced as `BigDecimal` which are then coerced as `Decimal` JS side.

We're now _only_ treating `Decimal` as `BigDecimal`, thus keeping `Float` columns as `Float`.


src/ast/index.rs
```diff
@@ -5,7 +5,7 @@ use super::{Column, Table};
 /// Used mainly for the transformation of a `INSERT` into a `MERGE`.
 #[derive(Debug, PartialEq, Clone)]
 pub enum IndexDefinition<'a> {
-    Single(Column<'a>),
+    Single(Box<Column<'a>>),
```
Weakky, line 8: clippy fix: https://rust-lang.github.io/rust-clippy/master/#large_enum_variant

rust-toolchain
```diff
@@ -1 +1 @@
-1.56.0
+1.57.0
```
Weakky, line 1: required by some dependency ; check the previous commit test failure if you need more info

PR #383, opened 2022-07-04
feat: support ordering nulls first & last

Weakky:
### ⚠️ Do not merge yet

First step to implement https://github.com/prisma/prisma/issues/4368

### DIFF
--- a/src/visitor/mssql.rs
+++ b/src/visitor/mssql.rs
@@ -618,6 +618,8 @@ impl<'a> Visitor<'a> for Mssql<'a> {
             }
         }
 
+        self.order_by_set = true;
+
         Ok(())
     }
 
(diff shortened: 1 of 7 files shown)


src/visitor/mssql.rs
```diff
@@ -549,14 +549,69 @@ impl<'a> Visitor<'a> for Mssql<'a> {
     fn visit_ordering(&mut self, ordering: Ordering<'a>) -> visitor::Result {
         let len = ordering.0.len();
 
-        for (i, (value, ordering)) in ordering.0.into_iter().enumerate() {
-            let direction = ordering.map(|dir| match dir {
-                Order::Asc => " ASC",
-                Order::Desc => " DESC",
-            });
+        fn render_ordering<'a>(visitor: &mut Mssql<'a>, direction: &str, value: Expression<'a>) -> visitor::Result {
```
dpetrick, line 552: I'm not a huge fan of inner functions, and would put them into the `impl` of the mssql visitor struct or into an extension trait. @pimeys what do you prefer as style for Quaint?

dpetrick (approved):
Nice work

src/visitor/mssql.rs
Weakky: I'm not a fan either, moved all inner functions to the `impl` block, it's much nicer anyway this way. Thanks 🙏🏻

PR #384, opened 2022-07-06
Allow setting Tx isolation on Tx start

dpetrick:
Allows setting an `IsolationLevel` when starting a new transaction. The underlying databases validate what levels of the ANSI + Snapshot levels are appropriate. Additionally, some databases require `SET` of the isolation level before a `BEGIN` of a transaction. 

Functional testing is thin, because it's hard to validate the isolation levels without either:
- Setting up dedicated scenarios for all databases to test.
- Checking that the connection executed the `SET` statement correctly.
- Somehow checking that the isolation level of the connection changed. Difficulty greatly depends on the database provider here. It's unclear if this is feasible for all providers.

### DIFF
--- a/src/connector/mssql.rs
+++ b/src/connector/mssql.rs
@@ -93,13 +93,17 @@ static SQL_SERVER_DEFAULT_ISOLATION: IsolationLevel = IsolationLevel::ReadCommit
 #[async_trait]
 impl TransactionCapable for Mssql {
     async fn start_transaction(&self, isolation: Option<IsolationLevel>) -> crate::Result<Transaction<'_>> {
-        // Isolation levels in SQL Server are set on the connection and live until they're unchanged
+        // Isolation levels in SQL Server are set on the connection and live until they're changed.
+        // Always explicitly setting the isolation level each time a tx is started (either to the given value
+        // or by using the default/connection string value) prevents transaction started on connection from
+        // the pool to have unexpected isolation levels set.
         Transaction::new(
             self,
             "BEGIN TRAN",
             isolation
                 .or_else(|| self.url.query_params.transaction_isolation_level)
                 .or_else(|| Some(SQL_SERVER_DEFAULT_ISOLATION)),
+            self.requires_isolation_first(),
         )
         .await
     }
@@ -430,6 +434,10 @@ impl Queryable for Mssql {
 
     fn begin_statement(&self) -> &'static str {
         "BEGIN TRAN"
+    }
+
+    fn requires_isolation_first(&self) -> bool {
+        true
     }
 }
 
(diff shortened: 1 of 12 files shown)


src/tests/query.rs
```diff
@@ -84,6 +84,60 @@ async fn transactions(api: &mut dyn TestApi) -> crate::Result<()> {
     Ok(())
 }
 
+#[test_each_connector(tags("mssql", "postgresql", "mysql"))]
+async fn transactions_with_isolation_works(api: &mut dyn TestApi) -> crate::Result<()> {
```
dpetrick, line 88: I don't have a great idea how to test that the TX isolation levels are set correctly, this is the best I came up with.

src/connector/mssql.rs
```diff
@@ -412,6 +376,11 @@ impl Queryable for Mssql {
         self.query_raw(sql, params).await
     }
 
+    async fn execute(&self, q: Query<'_>) -> crate::Result<u64> {
+        let (sql, params) = visitor::Mssql::build(q)?;
```
garrensmith, line 380: Why was this moved?

src/connector/mssql.rs
```diff
@@ -504,7 +484,12 @@ impl MssqlUrl {
         let transaction_isolation_level = props
             .remove("isolationlevel")
             .or_else(|| props.remove("isolation_level"))
-            .map(|level| IsolationLevel::from_str(&level))
+            .map(|level| {
```
garrensmith, line 487: This should be in its own smaller function.

pimeys (approved):

src/connector/mssql.rs
```diff
@@ -87,56 +88,24 @@ pub(crate) struct MssqlQueryParams {
...
 #[async_trait]
 impl TransactionCapable for Mssql {
-    async fn start_transaction(&self) -> crate::Result<Transaction<'_>> {
-        Transaction::new(self, "BEGIN TRAN").await
+    async fn start_transaction(&self, isolation: Option<IsolationLevel>) -> crate::Result<Transaction<'_>> {
+        // Isolation levels in SQL Server are set on the connection and live until they're changed.
+        // Always explicitly setting the isolation level each time a tx is started (either to the given value
+        // or by using the default/connection string value) prevents transactions started on connections from
+        // the pool to have unexpected isolation levels set.
+        Transaction::new(
```
garrensmith, line 100: I would change this to something like:
```
Transaction::new(
  self,
  "BEGIN TRAN",
  self.get_isolation_level(isolation)
);

```
That way you can put all the isolation comments in that function.

src/connector/transaction.rs
```diff
@@ -15,13 +19,31 @@ pub struct Transaction<'a> {
 }
 
 impl<'a> Transaction<'a> {
-    pub(crate) async fn new(inner: &'a dyn Queryable, begin_stmt: &str) -> crate::Result<Transaction<'a>> {
+    pub(crate) async fn new(
+        inner: &'a dyn Queryable,
+        begin_stmt: &str,
+        isolation_level: Option<IsolationLevel>,
+        isolation_first: bool,
```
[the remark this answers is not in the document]
Weakky: How about a `TransactionOptions` struct instead ?

Weakky (approved):

PR #387, opened 2022-07-14
Extend the expose-drivers cargo feature to SQLite and MSSQL

tomhoule:
For completeness.


src/connector/sqlite.rs
```diff
@@ -16,6 +16,11 @@ use tokio::sync::Mutex;
 
 pub(crate) const DEFAULT_SQLITE_SCHEMA_NAME: &str = "main";
 
+/// The underlying postgres driver. Only available with the `expose-drivers`
```
tomhoule, line 19: should be "sqlite" instead of "postgres"

garrensmith (approved):

PR #391, opened 2022-08-01
feat: add multiple functions & remove begins_with/ends_into

Weakky:
## Overview

Needed for https://github.com/prisma/prisma-engines/pull/3088

**BREAKING**:
- Removed begins_with, not_begins_with
- Removed ends_into, not_ends_into
- `.like` and `.not_like` now only renders `<left> LIKE <right>`. The consequence is:
	- `.like` should now be expressed as `.like(format!("%{}%", Value::text(...))` 
	- `.begins_with` should now be expressed as `.like(format!("{}%", Value::text(...))` 
	- `.ends_into` should now be expressed as `.like(format!("%{}", Value::text(...))` 
	- `.like` and `.not_like` can now reference other columns.

**New functions**:
- `concat()`
- `json_unquote()` (Postgres & MySQL ONLY)

**New comparable**:
- `.json_type_not_equals()` (Postgres & MySQL ONLY)	
- `.all()` (Postgres ONLY)
- `.any()` (Postgres ONLY)

**General improvements**:
- Added `JsonType::ColumnRef` to enable expressing `json_type(col) = json_type(col)`
- Fixed JSON equality on MariaDB (used to not work properly when one side of the equality was a function returning json)


src/ast/function/all_operator.rs
```diff
@@ -0,0 +1,28 @@
...
+/// # }
+/// ```
+pub fn all_operator<'a, T>(expr: T) -> Function<'a>
+where
+    T: Into<Expression<'a>>,
+{
+    let fun = AllOperator { expr: expr.into() };
+
+    fun.into()
+}
```
Weakky, line 28: So... `ALL` and `ANY` are apparently operators. I might sound dumb but this doesn't look like an operator to me... - `<left> operator <right>`
- `<left> operator`
- `operator <right>`

Those are all typical operator patterns as defined in most languages. But `ANY` and `ALL` have [this weird construct](https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-any/):

```
expresion operator ANY(subquery)
```

I'm not sure I get how `<left> <operator> <operator(ANY | ALL)> <right>` make any sense.

Anyway, because of that, I decided to treat `ANY` and `ALL` as functions, and not operator, which make it a lot easier to integrate with Quaint's assumptions of what an expression look like.

This is just to explain why I ended up with this very weird construct of `any_operator` and `all_operator` being functions 😭.

I'm open to suggestions, I know this is ugly

Weakky: Refactored ✅

garrensmith (approved):

garrensmith:
Nice work. Don't merge this to main until we are ready to merge the matching QE work.

PR #399, opened 2022-09-14
Wrap MySql subselects for update/delete

garrensmith:
MySql will error if a `Update` or `Delete` query has a subselect that references a table that is being updated or deleted to get around that, we need to wrap the table in a tmp table name

```
UPDATE `crabbywilderness` SET `val` = ?
WHERE (`crabbywilderness`.`id`)
IN (SELECT `t1`.`id` FROM `crabbywilderness` AS `t1` INNER JOIN `breakabletomatoes` AS `j` ON `j`.`id` = `t1`.`id2`)
```

### DIFF
--- a/src/tests/query.rs
+++ b/src/tests/query.rs
@@ -3479,18 +3479,10 @@ async fn update_with_subselect_using_main_table_does_not_throw_error(api: &mut d
 
     let id1 = Column::from((&table_1, "id"));
     let conditions = Row::from(vec![id1]).in_selection(selection);
-    let update = Update::table(&table_1).set("val", 1).so_that(conditions);
+    let update = Update::table(&table_1).set("val", 2).so_that(conditions);
 
-    use crate::{
-        ast::*,
-        visitor::{Mysql, Visitor},
-    };
-    let (sql, _) = Mysql::build(update.clone())?;
-
-    println!("SQL {:?}", sql);
-
-    let r = api.conn().update(update).await?;
-    println!("RES {:?}", r);
+    let res = api.conn().update(update).await?;
+    assert_eq!(res, 1);
 
     Ok(())
 }
--- a/src/tests/query.rs
+++ b/src/tests/query.rs
@@ -3474,8 +3474,8 @@ async fn update_with_subselect_using_main_table_does_not_throw_error(api: &mut d
     api.conn().insert(insert.into()).await?;
 
     let join = table_2.alias("j").on(("j", "id").equals(Column::from(("t1", "id2"))));
-    let a = table_1.clone().alias("t1");
-    let selection = Select::from_table(a).column(("t1", "id")).inner_join(join);
+    let t1_alias = table_1.clone().alias("t1");
+    let selection = Select::from_table(t1_alias).column(("t1", "id")).inner_join(join);
 
     let id1 = Column::from((&table_1, "id"));
     let conditions = Row::from(vec![id1]).in_selection(selection);
--- a/src/tests/query.rs
+++ b/src/tests/query.rs
@@ -3452,3 +3452,45 @@ async fn json_col_equal_json_col(api: &mut dyn TestApi) -> crate::Result<()> {
 
     Ok(())
 }
+
+#[test_each_connector(tags("mysql"))]
+async fn update_with_subselect_using_main_table_does_not_throw_error(api: &mut dyn TestApi) -> crate::Result<()> {
+    let table_1 = api.create_real_table("id int, id2 int, val int").await?;
+    let table_2 = api.create_real_table("id int").await?;
+
+    let insert = Insert::single_into(&table_1)
+        .value("id", 1)
+        .value("id2", 1)
+        .value("val", 1);
+    api.conn().insert(insert.into()).await?;
+
+    let insert = Insert::single_into(&table_1)
+        .value("id", 2)
+        .value("id2", 3)
+        .value("val", 1);
+    api.conn().insert(insert.into()).await?;
+
+    let insert = Insert::single_into(&table_2).value("id", 1);
+    api.conn().insert(insert.into()).await?;
+
+    let join = table_2.alias("j").on(("j", "id").equals(Column::from(("t1", "id2"))));
+    let a = table_1.clone().alias("t1");
+    let selection = Select::from_table(a).column(("t1", "id")).inner_join(join);
+
+    let id1 = Column::from((&table_1, "id"));
+    let conditions = Row::from(vec![id1]).in_selection(selection);
+    let update = Update::table(&table_1).set("val", 1).so_that(conditions);
+
+    use crate::{
+        ast::*,
+        visitor::{Mysql, Visitor},
+    };
+    let (sql, _) = Mysql::build(update.clone())?;
+
+    println!("SQL {:?}", sql);
+
+    let r = api.conn().update(update).await?;
+    println!("RES {:?}", r);
+
+    Ok(())
+}
--- a/src/tests/query.rs
+++ b/src/tests/query.rs
@@ -45,7 +45,7 @@ async fn aliased_null(api: &mut dyn TestApi) -> crate::Result<()> {
 
 #[test_each_connector]
 async fn select_star_from(api: &mut dyn TestApi) -> crate::Result<()> {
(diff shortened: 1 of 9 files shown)


tomhoule (approved):
It looks like a cleaner solution than the engine-based one we discussed on zoom :+1: I think @pimeys should have a look too because I'm not as familiar with the quaint query builder.

src/visitor/mysql.rs
```diff
@@ -247,6 +249,87 @@ impl<'a> Visitor<'a> for Mysql<'a> {
...
+    /// UPDATE `crabbywilderness` SET `val` = ?
+    /// WHERE (`crabbywilderness`.`id`)
+    /// IN (SELECT `t1`.`id` FROM `crabbywilderness` AS `t1`
+    /// INNER JOIN `breakabletomatoes` AS `j` ON `j`.`id` = `t1`.`id2`)
+    fn visit_sub_selection(&mut self, query: SelectQuery<'a>, _table: Option<Table<'a>>) -> visitor::Result {
+        match query {
+            SelectQuery::Select(select) => {
+                if let Some(table) = &self.table {
+                    if select.tables.contains(table) {
+                        let tmp_name = "tmp_subselect_table";
```
pimeys, line 319: Can we have name clashes with the outer or inner queries that might make some updates/deletes broken?

pimeys (commented):
Much better than the first round in the engines. There are a few places I'd consider a bit more before merging.

src/visitor/mysql.rs
```diff
@@ -247,6 +249,87 @@ impl<'a> Visitor<'a> for Mysql<'a> {
...
+
+                        return self.visit_select(sub_select);
+                    }
+                }
+
+                self.visit_select(*select)
+            }
+            SelectQuery::Union(union) => self.visit_union(*union),
+        }
+    }
```
Weakky, line 331: I don't understand where that `_table` argument is used. You're either passing `None` in the main `visitor.rs` and here you're using state from the `MySql` visitor that's being set when visiting an `update` or a `delete`. Am I missing something?

src/visitor/mysql.rs
```diff
@@ -12,6 +12,7 @@ use std::fmt::{self, Write};
 pub struct Mysql<'a> {
     query: String,
     parameters: Vec<Value<'a>>,
+    table: Option<Table<'a>>,
```
Weakky, line 15: Would you mind either being a bit more specific on the name of the field or just adding docs that explain what this is used for? `table` is pretty generic in the context of the visitor given that it's only set when visiting an update or a delete stmt.

src/visitor/mysql.rs
```diff
@@ -247,6 +249,87 @@ impl<'a> Visitor<'a> for Mysql<'a> {
         Ok(())
     }
 
+    // A walk through an `DELETE` statement for mysql
+    fn visit_delete(&mut self, delete: Delete<'a>) -> visitor::Result {
+        self.write("DELETE FROM ")?;
+        self.table = Some(delete.table.clone());
```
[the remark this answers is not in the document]
Weakky: An `Option` was likely used so that it could be initialized to something when the visitor is initialized

src/visitor/mysql.rs
garrensmith: sorry it was suppose to be removed.

src/visitor/mysql.rs
```diff
@@ -247,6 +249,87 @@ impl<'a> Visitor<'a> for Mysql<'a> {
...
+
+    /// MySql will error if a `Update` or `Delete` query has a subselect
+    /// that references a table that is being updated or deleted
+    /// to get around that, we need to wrap the table in a tmp table name
+    ///
+    /// UPDATE `crabbywilderness` SET `val` = ?
+    /// WHERE (`crabbywilderness`.`id`)
+    /// IN (SELECT `t1`.`id` FROM `crabbywilderness` AS `t1`
+    /// INNER JOIN `breakabletomatoes` AS `j` ON `j`.`id` = `t1`.`id2`)
+    fn visit_sub_selection(&mut self, query: SelectQuery<'a>, _table: Option<Table<'a>>) -> visitor::Result {
```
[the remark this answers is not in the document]
garrensmith: it was meant to be removed.

src/visitor/mysql.rs
garrensmith: I don't think so. We can only update one table at a time.

src/visitor/mysql.rs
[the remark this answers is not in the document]
garrensmith: Yes correct

src/tests/query.rs
```diff
@@ -3452,3 +3452,45 @@ async fn json_col_equal_json_col(api: &mut dyn TestApi) -> crate::Result<()> {
...
+    use crate::{
+        ast::*,
+        visitor::{Mysql, Visitor},
+    };
+    let (sql, _) = Mysql::build(update.clone())?;
+
+    println!("SQL {:?}", sql);
+
+    let r = api.conn().update(update).await?;
+    println!("RES {:?}", r);
```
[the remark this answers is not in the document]
garrensmith: I've changed it to check the result. All we really need to check is that it doesn't throw an error

src/visitor/mysql.rs
```diff
@@ -247,6 +249,87 @@ impl<'a> Visitor<'a> for Mysql<'a> {
...
+        if let Some(comment) = delete.comment {
+            self.write(" ")?;
+            self.visit_comment(comment)?;
+        }
+
+        Ok(())
+    }
+
+    /// A walk through an `UPDATE` statement for mysql
+    fn visit_update(&mut self, update: Update<'a>) -> visitor::Result {
```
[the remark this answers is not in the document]
garrensmith: removed

src/tests/test_api.rs
```diff
@@ -23,6 +23,7 @@ pub trait TestApi {
     async fn create_type_table(&mut self, r#type: &str) -> crate::Result<String>;
     fn render_create_table(&mut self, table_name: &str, columns: &str) -> (String, String);
     async fn create_table(&mut self, columns: &str) -> crate::Result<String>;
+    async fn create_real_table(&mut self, columns: &str) -> crate::Result<String>;
```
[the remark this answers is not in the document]
pimeys: Maybe name the other one as `create_temp_table` and this one `create_table`.

src/tests/test_api/mysql.rs
```diff
@@ -39,6 +39,17 @@ impl<'a> MySql<'a> {
 
         Ok(Self { names, conn, tag })
     }
+
+    fn render_perm_create_table(&mut self, table_name: &str, columns: &str) -> (String, String) {
```
pimeys, line 43: Do we btw. clean these tables after a test is run, so they don't leak to the other tests?

src/visitor/mysql.rs
```diff
@@ -12,6 +12,8 @@ use std::fmt::{self, Write};
 pub struct Mysql<'a> {
     query: String,
     parameters: Vec<Value<'a>>,
+    /// The table the query is acting on
```
pimeys, line 15:
```suggestion
    /// The table a deleting or updating query is acting on.
```

pimeys (commented):
Much better. Please write one more test to `visitor/mysql.rs` unit tests to show what kind of query gets generated from a certain input. Maybe one for delete and one for update. Then this is ready to be merged!

Weakky (approved):
Nothing else to add on top of what Julius already suggested in his latest review! Much better indeed, great work 👍

src/visitor/mysql.rs
```diff
@@ -895,4 +934,25 @@ mod tests {
             sql
         );
     }
+
+    #[test]
+    fn test_subselect_temp_table_wrapper() {
```
pimeys, line 939: Should we also have a test for deletes? Just copy this but make it to do `DELETE` so we also test that code path.

pimeys (approved):
One small test more, otherwise ready to go. Approving now, you decide is the test worth it.

src/visitor/mysql.rs
garrensmith: added

PR #400, opened 2022-09-19
feat(postgres): map 'std::io::ErrorKind::InvalidInput' coming from 'tokio_postgres' to 'quaint::error::ErrorKind::QueryInvalidInput'

jkomyno:
Context: https://www.notion.so/prismaio/Postgres-parameters-length-validation-ffc127b887984e02b9d931fcec4d6d15#62014932a75e438e9eb07306aee43224

Allows fixes to https://prisma/prisma/issues/8832 and https://prisma/prisma/issues/9326.

### DIFF
--- a/src/connector/postgres/error.rs
+++ b/src/connector/postgres/error.rs
@@ -349,19 +349,7 @@ fn try_extracting_io_error(err: &tokio_postgres::error::Error) -> Option<Error> 
 
     err.source()
         .and_then(|err| err.downcast_ref::<std::io::Error>())
-        .map(|err| {
-            let inner_io_error = Box::new(std::io::Error::new(err.kind(), format!("{}", err)));
-            match err.kind() {
-                // tokio_postgres will throw `std::io::Error`s with kind `InvalidInput` in several occasions,
-                // e.g., when the number of provided bind variables is >= 32768.
-                // We map these to `quaint::error::ErrorKind::QueryInvalidInput` to avoid confusing user-facing
-                // error messages.
-                std::io::ErrorKind::InvalidInput => ErrorKind::QueryInvalidInput(inner_io_error),
-
-                // fall back to a generic `ConnectionError`
-                _ => ErrorKind::ConnectionError(inner_io_error),
-            }
-        })
+        .map(|err| ErrorKind::ConnectionError(Box::new(std::io::Error::new(err.kind(), format!("{}", err)))))
         .map(|kind| Error::builder(kind).build())
 }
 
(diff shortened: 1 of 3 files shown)


src/connector/postgres/error.rs
```diff
@@ -335,7 +335,19 @@ fn try_extracting_io_error(err: &tokio_postgres::error::Error) -> Option<Error>
...
+            let inner_io_error = Box::new(std::io::Error::new(err.kind(), format!("{}", err)));
+            match err.kind() {
+                // tokio_postgres will throw `std::io::Error`s with kind `InvalidInput` in several occasions,
+                // e.g., when the number of provided bind variables is >= 32768.
+                // We map these to `quaint::error::ErrorKind::QueryInvalidInput` to avoid confusing user-facing
+                // error messages.
+                std::io::ErrorKind::InvalidInput => ErrorKind::QueryInvalidInput(inner_io_error),
+
+                // fall back to a generic `ConnectionError`
+                _ => ErrorKind::ConnectionError(inner_io_error),
```
janpio, line 348: Does this make sense as the generic fallback to `std::io::Error`? Isn't the danger that this covers still to many other cases, where we will keep having a unexpected "Can't connect ..." error message instead of anything helpful? Is there maybe a way to get at least the error message and include it as meta information?

jkomyno: The inner message is already included as "meta" information in `inner_io_error`. Currently, it gets lost elsewhere in the `query-engine`. I'll gladly fix that once this and [the companion PR](https://github.com/prisma/prisma-engines/pull/3201) are merged.

janpio: That would be amazing, as it would make bugs like the one you are fixing actually potentially debuggable.

pimeys:
A good start. Are you planning to catch this in engines?

pimeys (commented):
A good start. Are you planning to catch this in engines? What is the plan forward?

jkomyno:
> A good start. Are you planning to catch this in engines? What is the plan forward?

Thank you. The next step is reviewing and merging https://github.com/prisma/prisma-engines/pull/3201. It catches all `quaint::error::ErrorKind::QueryInvalidInput` and presents them as user-facing `DatabaseAssertionViolation` errors, which would forward the `tokio-driver` errors to the user (in the future, other assertion violation checks resulting in errors in other drivers can be mapped to this new user-facing error).

[Here](https://github.com/prisma/prisma/blob/e0a839db1bacffdac71222c0048c4d3f40525606/packages/client/tests/functional/issues/8832/tests.ts#L74-L78)'s an example of how the error would look like on the TS client side.
 
As you can see, it adds two new tests that currently succeed when run independently, but conflict on the CI (as stated [here](https://github.com/prisma/prisma-engines/pull/3201#issuecomment-1252300701)). We also mentioned the problem in this [internal thread](https://prisma-company.slack.com/archives/C4GCG53BP/p1663666458043209).

pimeys (approved):
Looks good now.

PR #403, opened 2022-09-19
Quote Returning on Sqlite pt2

do4gr:
Seems we have to work around a Sqlite bug/quirk:
https://sqlite.org/forum/info/6c141f151fa5c444db257eb4d95c302b70bfe5515901cf987e83ed8ebd434c49?t=h

### DIFF
--- a/src/visitor/sqlite.rs
+++ b/src/visitor/sqlite.rs
@@ -202,9 +202,9 @@ impl<'a> Visitor<'a> for Sqlite<'a> {
 
                 for (i, column) in returning.into_iter().enumerate() {
                     //yay https://sqlite.org/forum/info/6c141f151fa5c444db257eb4d95c302b70bfe5515901cf987e83ed8ebd434c49?t=h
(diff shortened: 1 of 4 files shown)


src/visitor/sqlite.rs
```diff
@@ -201,8 +201,10 @@ impl<'a> Visitor<'a> for Sqlite<'a> {
                 self.write(" RETURNING ")?;
 
                 for (i, column) in returning.into_iter().enumerate() {
+                    //yay https://sqlite.org/forum/info/6c141f151fa5c444db257eb4d95c302b70bfe5515901cf987e83ed8ebd434c49?t=h
+                    self.delimited_identifiers(&[&*column.name])?;
```
tomhoule, line 205: why do we use `delimited_identifiers` for a single identifier here?

tomhoule (approved):
just one question but lgtm :+1:

src/visitor/sqlite.rs
do4gr: Good point, refactored it.

PR #404, opened 2022-09-22
Map transaction write conflict/deadlock error codes to custom errors

SevInf:


src/tests/test_api.rs
```diff
@@ -34,5 +34,6 @@ pub trait TestApi {
     fn foreign_key(&mut self, parent_table: &str, parent_column: &str, child_column: &str) -> String;
     fn autogen_id(&self, name: &str) -> String;
     fn conn(&self) -> &crate::single::Quaint;
+    async fn new_conn(&self) -> crate::Result<crate::single::Quaint>;
```
pimeys, line 37: Even though it's a test setup, the naming here might be a bit better. It's not super clear what's the difference between this and `conn`...

pimeys (approved):
Super.

src/tests/test_api.rs
SevInf: Renamed it to `create_additional_connection`

PR #405, opened 2022-09-23
Wrap MSSQL RollbackWithoutBegin error

SevInf:
Tiberius PR: https://github.com/prisma/tiberius/pull/245

After fix in tiberius, MSSQL will throw this error if we try to rollback
the transaction which it have prevously rolled back as a result of
error.

Part of: https://github.com/prisma/prisma/issues/15607

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -127,12 +127,12 @@ features = ["bundled"]
 optional = true
 
 [target.'cfg(not(any(target_os = "macos", target_os = "ios")))'.dependencies.tiberius]
-version = "0.11.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -172,3 +172,6 @@ optional = true
 version = "0.6"
 features = ["compat"]
 optional = true
+
+[patch.crates-io]
```
SevInf, line 176: Note: remove before merging.

pimeys (approved):
Looks good to me.

PR #406, opened 2022-09-26
postgres: allow Value::Int in a doublearray

tomhoule:
We now accept converting to a double.

internal spec document: https://www.notion.so/prismaio/queryRaw-ints-in-float-arrays-18e5942570c04fd082727d749525f408


pimeys (approved):
If this only affects float arrays, let's do it.

PR #411, opened 2022-10-03
Add on Conflict update support for Postgres and Sqlite

garrensmith:
This adds support to do for `on conflict ... Update` for Postgres and Sqlite


src/visitor/postgres.rs
```diff
@@ -568,6 +576,26 @@ mod tests {
...
+
+        let query: Insert = Insert::single_into("users").value("foo", 10).into();
+
+        let query = query.on_conflict(OnConflict::Update(update, Vec::from(["foo".into()])));
+
+        let (sql, params) = Postgres::build(Insert::from(query).returning(vec!["foo"])).unwrap();
+
+        assert_eq!(expected.0, sql);
+        assert_eq!(expected.1, params);
+    }
```
Weakky, line 597: The same test for mysql would be useful

src/visitor/mysql.rs
```diff
@@ -251,6 +260,18 @@ impl<'a> Visitor<'a> for Mysql<'a> {
         Ok(())
     }
 
+    fn visit_upsert(&mut self, update: Update<'a>) -> visitor::Result {
+        if update.conditions.is_some() {
+            //TODO should this rather be an error or do we just ignore?
+            panic!("cannot use a where clause for a mysql upsert");
+        }
```
Weakky, line 267: Hmm, won't this limitation not fit with our existing upsert API? For instance, how would you express the following query given that MySQL doesn't support a `where` clause?

```
upsertOneUser({
  where: { email: "bob@prisma.io" },
  create: { id: 1, email: "alice@prisma.io", name: "alice" },
  data: { email: "alice@prisma.io" }
})
```

src/visitor/postgres.rs
garrensmith: I've removed mysql support for now

src/visitor.rs
```diff
@@ -326,6 +342,37 @@ pub trait Visitor<'a> {
...
+            self.write(" = ")?;
+            self.visit_expression(value)?;
+
+            if i < (len - 1) {
+                self.write(", ")?;
+            }
+        }
+
+        Ok(())
+    }
```
Weakky, line 374: Is there any difference between this and [`visit_update`](https://github.com/prisma/quaint/blob/34d170cc04c50cc8c4f3ac01d638720f404fb49b/src/visitor.rs#L295-L327) ?

Weakky: Ah, we don't visit the table in this upsert version, since we already visited it for the insert

garrensmith: Right. Its kind of annoying but I couldn't see another design

Weakky (approved):
LGTM 👍

PR #424, opened 2022-12-14
feat(PostgreSQL): support channel_binding query param

jannikkeye:
This PR adds support for the channel_binding query parameter for PostgreSQL.

I've added two tests to check for correct parsing as well as falling back to the default value if none or invalid value is present.

### DIFF
--- a/src/connector/postgres.rs
+++ b/src/connector/postgres.rs
@@ -21,7 +21,10 @@ use std::{
     sync::atomic::{AtomicBool, Ordering},
     time::Duration,
 };
-use tokio_postgres::{config::SslMode, Client, Config, Statement};
+use tokio_postgres::{
+    config::{ChannelBinding, SslMode},
+    Client, Config, Statement,
+};
 use url::Url;
 
 pub(crate) const DEFAULT_SCHEMA: &str = "public";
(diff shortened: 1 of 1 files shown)


src/connector/postgres.rs
```diff
@@ -403,6 +411,19 @@ impl PostgresUrl {
...
                     application_name = Some(v.to_string());
                 }
+                "channel_binding" => {
+                    match v.as_ref() {
+                        "disable" => channel_binding = ChannelBinding::Disable,
+                        "prefer" => channel_binding = ChannelBinding::Prefer,
+                        "require" => channel_binding = ChannelBinding::Require,
+                        _ => {
+                            tracing::debug!(
+                                message = "Unsupported Channel Binding channel_binding, defaulting to `prefer`",
```
pimeys, line 421:
```suggestion
                                message = "Unsupported Channel Binding {channel_binding}, defaulting to `prefer`",
```

pimeys (changes requested):

pimeys:
a small nitpick, otherwise ready to go

pimeys:
thanks.

pimeys:
It needs testing and documentation. Which is not scheduled for the current sprint. Would you like to do it?

pimeys:
And quaint should be updated in the engines repo.

jannikkeye:
> It needs testing and documentation. Which is not scheduled for the current sprint. Would you like to do it?

Yeah, sure I’ll add the docs, I think I know where, have added a param before.

PR #431, opened 2023-01-27
Emit query log in quaint:query span, rather than in its parent

miguelff:
Tracked in https://github.com/prisma/client-planning/issues/158

Addressed the problem expressed in https://github.com/prisma/client-planning/issues/158#issuecomment-1405530608

Before this change query log events where not attached to the `quaint:query` span but its parent. This had implications on log capturing. mainly, that `BEGIN` iTX queries where not logged until its owner span (itx_runner) was not closed at the end of the transaction.

### DIFF
--- a/src/connector/metrics.rs
+++ b/src/connector/metrics.rs
@@ -14,8 +14,16 @@ where
     U: Future<Output = crate::Result<T>>,
 {
     let span = info_span!("quaint:query", "db.statement" = %query);
+    do_query(tag, query, params, f).instrument(span).await
+}
+
+async fn do_query<'a, F, T, U>(tag: &'static str, query: &'a str, params: &'a [Value<'_>], f: F) -> crate::Result<T>
+where
+    F: FnOnce() -> U + 'a,
+    U: Future<Output = crate::Result<T>>,
+{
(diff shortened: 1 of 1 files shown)


src/connector/metrics.rs
```diff
@@ -14,8 +14,16 @@ where
     U: Future<Output = crate::Result<T>>,
 {
     let span = info_span!("quaint:query", "db.statement" = %query);
```
miguelff, line 16: In the query engine binary and node API, This span is [filtered-in by target](https://github.com/prisma/prisma-engines/blob/61863c7b1810c243ede42ae8ed40dc849b38c193/query-engine/query-engine-node-api/src/logger.rs#L35) , because it is user facing but doesn't have a `user_facing` attribute. Should we add the attribute to prevent this additional filtering based on the target? 


I know, it's coupling quiant to engine needs. But isn't that already based on all the tracing features added to it?

miguelff:
Given the [correct behavior](https://github.com/prisma/prisma-engines/pull/3618#issuecomment-1406638266) of this patch, the simplicity of the change, and the lack of required reviewers for quaint, I will merge unless there's strong opposition.

PR #434, opened 2023-02-21
We don't need to use our fork of mysql anymore

pimeys:
Just defining the features for tls crates in quaint is enough.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,8 +44,8 @@ all = [
 ]
 
 vendored-openssl = [
-  "postgres-native-tls/vendored-openssl",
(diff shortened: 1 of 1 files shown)


tomhoule (commented):
Like we discussed in DMs, I'm worried about the potential for this silently breaking. Do you feel comfortable like this, or should we go with cargo deny and a strict rule for tokio-native-tls? I think we'll want cargo deny for one reason or another, might as well start now.

pimeys:
Cargy deny wants us to separately list all skipped crates from the checks, which is just way too much work for everybody. We decided on Slack to look into dupes for tls stuff in the PR reviews and merge this now.

PR #438, opened 2023-03-31
tech-debt: use upstream mobc + small fixes

miguelff:
Tracked in https://github.com/prisma/client-planning/issues/312 (private)

Necessary for https://github.com/prisma/prisma-engines/pull/3833

Updates quaint to use upstream mobc, and connection-string 0.2, which doesn't cause conflicts in prisma-engines, like the last patch of 0.1 did.


tomhoule (approved):
lgtm :+1:

PR #441, opened 2023-04-05
perf(pg): remove one roundtrip on connection

Weakky:
## Overview

Removes one roundtrip to set the `client-encoding` and `search_path`. Both can be set through client options.

- [`client_encoding`](https://www.postgresql.org/docs/8.0/multibyte.html#:~:text=Using%20the%20configuration%20variable%20client_encoding.%20If%20the%20client_encoding%20variable%20is%20set%2C%20that%20client%20encoding%20is%20automatically%20selected%20when%20a%20connection%20to%20the%20server%20is%20made.%20(This%20can%20subsequently%20be%20overridden%20using%20any%20of%20the%20other%20methods%20mentioned%20above.))
- [`search_path`](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-SEARCH-PATH)


src/connector/postgres.rs
```diff
@@ -547,20 +551,6 @@ impl PostgreSql {
...
         }));
 
-        // SET NAMES sets the client text encoding. It needs to be explicitly set for automatic
-        // conversion to and from UTF-8 to happen server-side.
-        //
-        // Relevant docs: https://www.postgresql.org/docs/current/multibyte.html
-        let session_variables = format!(
-            r##"
-            {set_search_path}
-            SET NAMES 'UTF8';
```
Weakky, line 557: `rust-postgres` already sets that to `UTF8`, as you can see here https://github.com/prisma/rust-postgres/blob/614e66c539c6127a8749225d4a6f7c67fe21c0bb/tokio-postgres/src/connect_raw.rs#L120.

If you're unsure whether it is referring to the same setting, [have a look here](https://www.postgresql.org/docs/current/multibyte.html#:~:text=Using%20the%20configuration%20variable%20client_encoding.%20If%20the%20client_encoding%20variable%20is%20set%2C%20that%20client%20encoding%20is%20automatically%20selected%20when%20a%20connection%20to%20the%20server%20is%20made.%20(This%20can%20subsequently%20be%20overridden%20using%20any%20of%20the%20other%20methods%20mentioned%20above.))

src/connector/postgres.rs
```diff
@@ -486,7 +486,11 @@ impl PostgresUrl {
 
         if let Some(connect_timeout) = self.query_params.connect_timeout {
             config.connect_timeout(connect_timeout);
-        };
+        }
+
+        if let Some(schema) = &self.query_params.schema {
+            config.search_path(schema.clone());
+        }
```
tomhoule, line 493: I see we move things out of query_params just above. Could we avoid the clone here?

Weakky: Things above can be moved because they implement `Copy`. Whether it's `Duration`, `SslMode`, `ChannelBinding` or `bool`

tomhoule: Ach, I would have expected that to be in a `From` impl or otherwise take self by value, but indeed, it's not the case.

PR #444, opened 2023-04-06
Add to_vec to a value

pimeys:


janpio:
Done via https://github.com/prisma/quaint/pull/446/files

PR #448, opened 2023-04-14
feat: optimize search_path network roundtrip

Weakky:
## Overview

Avoids a network roundtrip at connection to set the search path when possible.

We decide whether we do that roundtrip based on the postgres flavour set in the url.

The different behaviours are:
- Postgres: Always avoid a network roundtrip by setting the search path through client connection parameters.
- Cockroach: Avoid a network roundtrip if the schema name is deemed "safe" (i.e. no escape quoting required). Otherwise, set the search path through a database query. We do that because Cockroach has issues when parsing schema names https://github.com/cockroachdb/cockroach/issues/101394
- Unknown: Always add a network roundtrip by setting the search path through a database query. This ensures backward compatibility for the Migration Engine. The default flavour is set to Unknown, thus always setting the search_path through a database query which ensures that it works fine both on PG & CRDB.

### DIFF
--- a/src/connector/postgres.rs
+++ b/src/connector/postgres.rs
@@ -147,6 +147,36 @@ impl SslParams {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub enum PostgresFlavour {
+    Postgres,
+    Cockroach,
+    Unknown,
+}
+
+impl PostgresFlavour {
+    /// Returns `true` if the postgres flavour is [`Postgres`].
+    ///
+    /// [`Postgres`]: PostgresFlavour::Postgres
+    fn is_postgres(&self) -> bool {
+        matches!(self, Self::Postgres)
+    }
+
+    /// Returns `true` if the postgres flavour is [`Cockroach`].
+    ///
+    /// [`Cockroach`]: PostgresFlavour::Cockroach
+    fn is_cockroach(&self) -> bool {
+        matches!(self, Self::Cockroach)
+    }
+
+    /// Returns `true` if the postgres flavour is [`Unknown`].
+    ///
+    /// [`Unknown`]: PostgresFlavour::Unknown
+    fn is_unknown(&self) -> bool {
+        matches!(self, Self::Unknown)
+    }
+}
+
 /// Wraps a connection url and exposes the parsing logic used by Quaint,
 /// including default values.
 #[derive(Debug, Clone)]
@@ -154,6 +184,7 @@ impl SslParams {
 pub struct PostgresUrl {
     url: Url,
     query_params: PostgresUrlQueryParams,
+    flavour: PostgresFlavour,
 }
 
 impl PostgresUrl {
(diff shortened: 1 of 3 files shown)


src/connector/postgres.rs
```diff
@@ -827,10 +915,40 @@ impl Queryable for PostgreSql {
     }
 }
 
+/// Returns true if a Postgres identifier is considered "safe".
+///
+/// In this context, "safe" means that the value of an identifier would be the same quoted and unquoted. In other words, that it does _not_ need to be quoted.
+///
+/// Spec can be found here: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
+fn is_safe_identifier(ident: &str) -> bool {
```
pimeys, line 923: What about identifiers that are reserved words? For example, postgresql (or CDB) would not like if you have a schema called `user` or `if` and use it in a query without quoting...

src/connector/postgres.rs
```diff
@@ -547,19 +610,20 @@ impl PostgreSql {
...
         }));
 
-        // SET NAMES sets the client text encoding. It needs to be explicitly set for automatic
-        // conversion to and from UTF-8 to happen server-side.
-        //
-        // Relevant docs: https://www.postgresql.org/docs/current/multibyte.html
-        let session_variables = format!(
-            r##"
-            {set_search_path}
-            SET NAMES 'UTF8';
```
janpio, line 557: Not needed at all?

src/connector/postgres.rs
```diff
@@ -641,6 +705,30 @@ impl PostgreSql {
...
 }
 
+// A SearchPath connection parameter (Display-impl) for connection initialization.
+struct CockroachSearchPath<'a>(&'a str);
+
+impl Display for CockroachSearchPath<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.0)?;
+
+        Ok(())
```
tomhoule, line 715: The body can be simplified to just `f.write_str(self.0)`

src/connector/postgres.rs
tomhoule: Hmm good point. Is there a convenient list of the reserved keywords we could use? Is it large? Complicated.

tomhoule (approved):
Not sure what to do about @pimeys 's comment about reserved keywords. It's an edge case but it's a very valid point.

pimeys:
> Not sure what to do about @pimeys 's comment about reserved keywords. It's an edge case but it's a very valid point.

`user` is even a very common name on the databases...

src/connector/postgres.rs
Weakky: Are you specifically referring to `SET NAMES UTF8`? It is already set by `tokio-postgres` via connection parameters https://github.com/sfackler/rust-postgres/blob/65c1d146a6b3b91652f211d8c72e776164ee87a4/tokio-postgres/src/connect_raw.rs#L114

PR #450, opened 2023-04-18
Back to mainline mysql-async:

pimeys:
We have everything merged now, and we can just set the openssl flags to the libraries in quaint, and mysql_async will pick them up.


miguelff (approved):
Awesome!

PR #453, opened 2023-04-20
fix: set search path via query when pgbouncer is on

Weakky:
## Overview

- PGBouncer unfortunately does not support the `search_path` connection parameter. So we set it via a query (as we used to) when pgbouncer=true.

### DIFF
--- a/src/connector/postgres.rs
+++ b/src/connector/postgres.rs
@@ -512,6 +512,27 @@ impl PostgresUrl {
         self.query_params.connection_limit
     }
 
+    /// On Postgres, we set the SEARCH_PATH and client-encoding through client connection parameters to save a network roundtrip on connection.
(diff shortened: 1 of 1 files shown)


src/connector/postgres.rs
```diff
@@ -1457,4 +1472,63 @@ mod tests {
...
+
+        let mut pg_url = PostgresUrl::new(url).unwrap();
+        pg_url.set_flavour(PostgresFlavour::Postgres);
+
+        let config = pg_url.to_config();
+
+        // PGBouncer does not support the `search_path` connection parameter.
+        // When `pgbouncer=true`, config.search_path should be None,
+        // And the `search_path` should be set via a db query after connection.
+        assert_eq!(config.get_search_path(), None);
```
Weakky, line 1490: Added a bunch of other unit tests, just in case.

miguelff (approved):

